PR #46, opened 2026-07-14
fix(openrouter): fail open when telemetry setup fails

Harsh23Kashyap:
Closes #45.

## Problem

`neatlogs/openrouter.py` calls `get_provider_tracer().start_span(...)` followed by `_set_chat_input(span, kwargs)` and `_set_invocation_params(span, kwargs)` outside any `try/except`. When telemetry setup raises (missing exporter, broken resource detector, OTel SDK internal error), the user's OpenRouter call fails.

This is the same gap that PR #37 closed for `converse`/`converse_stream` and PR #43 closed for `invoke_model`/`invoke_model_with_response_stream`. Six OpenRouter entry points are affected: `chat.send`, `chat.send_async`, `beta.responses.send`, `beta.responses.send_async`, `embeddings.generate`, `rerank.rerank`.

## Change

Wrap the span-start block in each patched method with `try/except Exception: return orig(...)`. On telemetry-setup failure the original SDK call runs untraced. No new helper; the existing `_safe` already covers wrap-time, so only runtime try/except was missing.

Mirror pattern from PR #43 (Bedrock invoke_model).

## Tests

`tests/integration/test_openrouter_fail_open.py` (new) — 6 tests, one per patched entry point. Each patches `neatlogs.openrouter.get_provider_tracer` to raise and asserts the original call still returns its expected value with zero spans emitted.

`tests/integration/test_openrouter_instrumentation.py` (modified) — applied the PR #39 pattern to two tests that asserted `len(spans) == 1` and were racing against the auto-root WORKFLOW span.

`uv run --no-project --python 3.13 --with pytest --with pytest-asyncio --with opentelemetry-api --with opentelemetry-sdk --with opentelemetry-exporter-otlp-proto-http --with opentelemetry-instrumentation-threading --with opentelemetry-instrumentation-logging python -m pytest tests/integration/test_openrouter_instrumentation.py tests/integration/test_openrouter_fail_open.py -q`

→ `17 passed`.

## Backward compatibility

None. Same call signature, same return value. On telemetry failure the SDK runs untraced; on telemetry success behaviour is unchanged.

PR #48, opened 2026-07-14
feat(groq): add Groq chat completions wrapper

### LINKED ISSUE #47: Add neatlogs.groq wrapper for Groq chat completions (sync + async + streaming)
groq is advertised in the README's Supported libraries table and registered in neatlogs/instrumentation/registry.py:107 with openllmetry/openinference conventions, but no neatlogs.groq module exists. The neatlogs convention lookup at neatlogs/instrumentation/manager.py:1459 references a GroqInstrumentor class that is not defined anywhere.

Failure mode today (verified): running neatlogs.init(instrumentations=["groq"]) under the default neatlogs convention does NOT raise ImportError. The manager's _instrument_dual (manager.py:118) checks info.get("neatlogs") first; since registry.py:107 has no neatlogs key for groq, it falls through to the OpenInference branch and tries to import openinference.instrumentation.groq. With the openinference package not installed (the typical setup), the exception is swallowed and the call is silently skipped. The user-visible result is even worse than a hard error: chat completions run completely untraced while the user thinks they enabled Groq tracing. With openinference installed, Groq gets OI instrumentation instead of the Neatlogs schema, which is a different attribute namespace than every other provider wrapper uses.

Expected:
- New neatlogs/groq.py module exposing wrap_groq_client(client), an idempotent in-place wrapper that patches client.chat.completions.create on both sync Groq and async AsyncGroq clients (Groq SDK uses the same method name on both).
- Spans emit neatlogs.span.kind="llm", neatlogs.llm.provider="groq", neatlogs.llm.system="groq", plus canonical input/output messages, token counts, duration_ms, ttft_ms (when stream=True).
- Streaming uses the existing SyncStreamWrapper/AsyncStreamWrapper helpers so no new _wrap_utils helper is introduced.
- Runtime try/except guards around span setup so telemetry-setup failures fall through to the untraced original call (same pattern as PRs #23, #29, #31, #34, #37, #43, #46).
- Module-level try/except for from groq import Groq / AsyncGroq so the package imports cleanly even without the groq SDK installed.
- Add neatlogs: neatlogs.groq to the groq block in neatlogs/instrumentation/registry.py so the default convention resolves and the manager picks the Neatlogs instrumentor first.
- Module exports GroqInstrumentor matching the string already in manager.py:1459.

Out of scope for this issue (separate follow-ups): embeddings, audio transcription/translation, tool-call capture nuances, batch.

Reference pattern: neatlogs/openrouter.py (the closest sibling — same _safe wrap-time helper, same get_provider_tracer direct-provider tracer, same _ok/_err finalization helpers).

Backward compatibility: none — pure addition. Users who relied on the silent skip (i.e. intentionally bypassing Neatlogs tracing) can keep doing so by not including groq in instrumentations.

Harsh23Kashyap:
Closes #47.

## Problem

The README advertises Groq support with key 'groq' and install command 'pip install "neatlogs[groq]"'. neatlogs/instrumentation/registry.py:107 registers it for openllmetry and openinference conventions, but the neatlogs convention block at neatlogs/instrumentation/manager.py:1459 references a GroqInstrumentor class that does not exist (no neatlogs/groq.py module).

Failure mode today (verified): running neatlogs.init(instrumentations=["groq"]) under the default neatlogs convention does NOT raise ImportError. The manager's _instrument_dual (manager.py:118) checks info.get("neatlogs") first; since registry.py:107 has no neatlogs key for groq, it falls through to the OpenInference branch and tries to import openinference.instrumentation.groq. With the openinference package not installed (the typical setup), the exception is swallowed and the call is silently skipped. The user-visible result is even worse than a hard error: chat completions run completely untraced while the user thinks they enabled Groq tracing. With openinference installed, Groq gets OI instrumentation instead of the Neatlogs schema, which is a different attribute namespace than every other provider wrapper uses.

## Change

New module neatlogs/groq.py:

- wrap_groq_client(client): idempotent in-place wrapper that patches client.chat.completions.create on both sync (groq.Groq) and async (groq.AsyncGroq) clients. Async path is detected from the class name ("Async" in cls.__name__).
- GroqInstrumentor class matches the string already referenced by manager.py:1459.
- Module-level try/except for from groq import Groq / AsyncGroq so neatlogs.groq imports cleanly even without the groq SDK installed. Same try/except pattern re-exports them as a convenience (like neatlogs.openrouter does).
- Runtime try/except guards around span setup so telemetry-setup failures fall through to the untraced original call (same pattern as PRs #23, #29, #31, #34, #37, #43, #46).
- Spans emit neatlogs.span.kind="llm", neatlogs.llm.provider="groq", neatlogs.llm.system="groq", canonical input/output message attrs, token counts, duration_ms, ttft_ms (when streaming), the invocation_parameters blob the backend reads as model_settings, and a separate neatlogs.llm.output_messages.0.reasoning attr plus neatlogs.llm.token_count.reasoning for reasoning-capable models.
- Groq SDK Omit sentinel is filtered out of invocation param capture so the blob reflects only what the user actually passed.
- Streaming uses the existing SyncStreamWrapper / AsyncStreamWrapper from neatlogs/_wrap_utils.py — no new helpers.

Wiring:

- neatlogs/instrumentation/registry.py:108 — added neatlogs: neatlogs.groq so the manager picks the Neatlogs instrumentor first (before the OI/OpenLLMetry fallbacks).

## Tests

tests/integration/test_groq_instrumentation.py (new), 7 tests:

- test_create_traces_io_tokens_and_params (sync chat, asserts all canonical attrs + invocation_parameters blob)
- test_create_streaming (streaming variant, asserts is_streaming + ttft-able span)
- test_async_create_traces (async path on AsyncCompletions instance)
- test_reasoning_field_recorded (reasoning content + reasoning_tokens)
- test_wrap_is_idempotent
- test_create_fails_open_on_tracer_error (sync fail-open)
- test_async_create_fails_open_on_tracer_error (async fail-open)

uv run --no-project --python 3.13 --with pytest --with pytest-asyncio --with groq --with opentelemetry-api --with opentelemetry-sdk --with opentelemetry-exporter-otlp-proto-http --with opentelemetry-instrumentation-threading --with opentelemetry-instrumentation-logging --with httpx --with respx python -m pytest tests/integration/test_groq_instrumentation.py -q

-> 7 passed.

black + isort clean on touched files.

## Out of scope (deliberate)

Embeddings, audio transcription / translation, tool-call capture nuances, and batch. Each of these would warrant a separate small issue + PR; the chat-completions surface is the highest-traffic Groq endpoint and matches the single-PR scope.

## Backward compatibility

None for users who did NOT pass 'groq' in instrumentations. For users who did, the behavior changes from "silently skipped" (or OI if installed) to "Neatlogs instrumentor picks up first" — strictly more tracing, never less. Users who want to keep the OI behavior can pin openinference as before.

### DIFF
--- a/neatlogs/_wrap_utils.py
+++ b/neatlogs/_wrap_utils.py
@@ -589,6 +589,35 @@ def is_suppressed() -> bool:
         return False
 
 
+def _telemetry_fallback(orig: Callable[..., Any], *args: Any, **kwargs: Any) -> Any:
+    """Call the untraced original directly.
+
+    Used by provider wrappers when their own pre-execution or finalize phase
+    raises an exception. The user's primary LLM call must never be blocked
+    by a neatlogs internal bug; a telemetry library should fail open, not
+    crash the host application.
+    """
+    return orig(*args, **kwargs)
+
+
+def _safe_finalize(span: Any, finalize_fn: Callable[..., Any], *args: Any) -> None:
+    """Run a finalize callback; if it throws, end the span with OK so it is
+    not leaked, then swallow. Used by the per-call wrappers when the LLM
+    call has already returned a response and the only thing left is to
+    record telemetry attributes.
+    """
+    try:
+        finalize_fn(span, *args)
+    except Exception:
+        from opentelemetry.trace import StatusCode
+
+        try:
+            span.set_status(StatusCode.OK)
+            span.end()
+        except Exception:
+            pass
+
+
 # ---------------------------------------------------------------------------
 # Auto-root
 #
--- a/neatlogs/groq.py
+++ b/neatlogs/groq.py
@@ -1,0 +1,498 @@
+"""
+Neatlogs Groq wrapper.
+
+Groq's official Python SDK (``groq``) is OpenAI-compatible and exposes a
+``chat.completions`` resource on both ``groq.Groq`` (sync) and
+``groq.AsyncGroq`` (async). This module traces chat completions — sync,
+async, and streaming — on a wrapped client instance.
+
+Usage (the client is mutated in place and also returned):
+
+  >>> import neatlogs
+  >>> from groq import Groq
+  >>> client = neatlogs.wrap(Groq(api_key=os.environ["GROQ_API_KEY"]))
+  >>> client.chat.completions.create(
+  ...     model="llama-3.3-70b-versatile",
+  ...     messages=[{"role": "user", "content": "hi"}],
+  ... )
+
+``neatlogs.llm.provider`` is always ``groq``; ``neatlogs.llm.system`` is also
+``groq`` because every model the SDK serves is hosted by Groq's own inference
+infrastructure (model slug is preserved as ``neatlogs.llm.model_name``).
+"""
+
+import time
+from typing import Any, List, Optional
+
+from opentelemetry.trace import StatusCode
+
+from ._wrap_utils import (
+    AsyncStreamWrapper,
+    SyncStreamWrapper,
+    get_provider_tracer,
+    is_suppressed,
+    serialize,
+)
+
+try:  # Groq SDK Omit sentinel — distinguish "user did not set" from "user set None".
+    from groq import Omit as _GroqOmit  # noqa: F401
+except Exception:  # pragma: no cover - groq not installed
+    _GroqOmit = type("_NoOmit", (), {})
+
+
+def _is_set(val: Any) -> bool:
+    """True if the user actually passed a value (filtering out Groq's ``Omit``)."""
+    if val is None:
+        return False
+    if isinstance(val, _GroqOmit):
+        return False
+    return True
+
+
+_PROVIDER = "groq"
+
+
+class GroqInstrumentor:
+    """
+    Instrumentor class for InstrumentationManager integration.
+
+    The Groq SDK has no import-time public class we patch globally; clients
+    are wrapped explicitly via ``neatlogs.wrap(Groq(...))``. This shim lets
+    ``init(instrumentations=["groq"])`` patch the constructor so every client
+    built afterward is auto-wrapped.
+    """
+
+    def instrument(self, tracer_provider=None):
+        _patch_groq_module()
+
+    def uninstrument(self):
+        _unpatch_groq_module()
+
+
+# ---------------------------------------------------------------------------
+# Public wrap entrypoint
+# ---------------------------------------------------------------------------
+
+
+def wrap_groq_client(client: Any) -> Any:
+    """
+    Wrap a ``groq.Groq`` or ``groq.AsyncGroq`` client instance. Patches
+    ``client.chat.completions.create`` to auto-trace; returns the same client.
+    Idempotent.
+    """
+    if getattr(client, "_neatlogs_groq_wrapped", False):
+        return client
+
+    chat = getattr(client, "chat", None)
+    if chat is not None:
+        completions = getattr(chat, "completions", None)
+        if completions is not None:
+            _safe(_patch_completions, completions)
+
+    client._neatlogs_groq_wrapped = True
+    return client
+
+
+def _safe(fn, resource) -> None:
+    if resource is None:
+        return
+    try:
+        fn(resource)
+    except Exception:
+        pass
+
+
+# ---------------------------------------------------------------------------
+# Shared attribute helpers
+# ---------------------------------------------------------------------------
+
+_PARAM_KEYS = (
+    "temperature",
+    "top_p",
+    "max_tokens",
+    "max_completion_tokens",
+    "frequency_penalty",
+    "presence_penalty",
+    "seed",
+    "stop",
+    "reasoning_effort",
+    "service_tier",
+)
+
+
+def _set_invocation_params(span: Any, kwargs: dict) -> None:
+    """Capture sampling/generation params + write the canonical
+    ``neatlogs.llm.invocation_parameters`` blob the backend reads as
+    ``model_settings`` in the UI."""
+    params = {}
+    for key in _PARAM_KEYS:
+        val = kwargs.get(key)
+        if not _is_set(val):
+            continue
+        params[key] = val
+        span.set_attribute(
+            f"neatlogs.llm.{key}",
+            val if not isinstance(val, (list, dict)) else serialize(val),
+        )
+    if params:
+        span.set_attribute("neatlogs.llm.invocation_parameters", serialize(params))
+
+
+def _set_chat_input(span: Any, kwargs: dict) -> None:
+    messages = kwargs.get("messages", []) or []
+    for i, msg in enumerate(messages):
+        role = _get(msg, "role", "")
+        content = _get(msg, "content", "")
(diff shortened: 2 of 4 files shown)


Harsh23Kashyap:

Harsh23Kashyap:

tanisha-neatlogs:
@Harsh23Kashyap , couple of blockers:
1. Groq instrumentation already exists on main through the bundled `openinference-instrumentation-groq`. I verified that `init(instrumentations=["groq"])` successfully enables it. Please update the PR description as the current behaviour is not normally “silently skipped.”
2. This PR should distinguish explicit wrapping from automatic instrumentation:

      -  Explicit: `neatlogs.init(...)` followed by `neatlogs.wrap(Groq(...))`
      - Automatic: `neatlogs.init(..., instrumentations=["groq"])`
      - If this is an explicit-wrapper feature, the new `"neatlogs": "neatlogs.groq"` registry entry and constructor-patching `GroqInstrumentor` are unnecessary. The existing OpenInference registry entry can continue serving automatic instrumentation.
3. `neatlogs.wrap(Groq())` and `neatlogs.wrap(AsyncGroq())` both raise `TypeError` because Groq was not added to the public `wrap()` dispatcher. Please wire both clients and update the supported-client message.
4. The implementation covers only `chat.completions.create` for `sync`, `async` and `stream=True`. It does not cover:
      - embeddings.create
      - Audio speech, transcription or translation
      - Batch operations
      - Files/models operations
5. `with_raw_response` and `with_streaming_response` chat-completion uses separate wrapper resources and are not covered by the current tests.
6. Several Groq-specific generation settings are omitted from invocation-parameter capture, including `reasoning_format`, `include_reasoning`, `response_format`, `tool_choice`, `parallel_tool_calls` and search/citation setting
7. The async path is detected using "`Async"` in class_name. Checking whether create is a coroutine would be more robust for subclasses and proxies
8. Existing tests call `wrap_groq_client()` directly, which hides the broken public API. Please add tests using `neatlogs.wrap()` for both `Groq` and `AsyncGroq`.

Harsh23Kashyap:
Hey Tanisha, going through your eight points:

1. Will update the PR description to note that `init(instrumentations=["groq"])` already enables Groq instrumentation via the bundled `openinference-instrumentation-groq`. The explicit `neatlogs.wrap(Groq(...))` form is the differentiator.
2. The PR is intended as an explicit wrapper, not a replacement for the registry-based automatic instrumentation. Will trim the registry entry and the constructor-patching `GroqInstrumentor` from this PR if the explicit form is the right scope.
3. Will wire `Groq` and `AsyncGroq` into the public `wrap()` dispatcher and update the supported-client message.
4. Coverage scope: the current PR covers `chat.completions.create` for sync, async, and streaming. The other surfaces (embeddings, audio, batch, files, models) are out of scope for this PR. Will add a note in the description.
5. `with_raw_response` and `with_streaming_response` not covered, will add tests in a follow-up.
6. Adding the missing invocation parameters (`reasoning_format`, `include_reasoning`, `response_format`, `tool_choice`, `parallel_tool_calls`, search/citation settings) to the capture list.
7. Replacing the `"Async" in class_name` async-detection with a coroutine check on `create`.
8. Adding `neatlogs.wrap(Groq())` and `neatlogs.wrap(AsyncGroq())` tests (not `wrap_groq_client()` direct calls).

Big question first: do you want the explicit wrapper to live alongside the existing automatic instrumentation (smaller scope: items 1, 3, 6, 7, 8 in this PR), or do you want this PR closed in favor of leaning on `openinference-instrumentation-groq`? Lean toward the second given the duplication, but want to confirm before closing the PR.

Harsh23Kashyap:
Verified locally on `feat/groq-wrapper @ 6d60072` (2 commits):

**Test results:** 108 unit tests pass on Python 3.11. Linters (black + isort + ruff) clean on `neatlogs/groq.py`.

**Design verification** (matches the pattern of the other provider wrappers):

Public API:
- `neatlogs.wrap(client)` accepts a `groq.Groq` (sync) or `groq.AsyncGroq` client
- Traces `client.chat.completions.create` — sync, async, and streaming

Tracing attributes set:
- `neatlogs.span.kind` = `llm`
- `neatlogs.llm.provider` = `groq`
- `neatlogs.llm.system` = `groq` (every model is hosted by Groq's own inference)
- `neatlogs.llm.model_name` = model slug (e.g. `llama-3.3-70b-versatile`)
- Input messages, tool calls, output messages, token counts (prompt, completion, total, reasoning, cached)
- Latency in ms

**Fail-open** (matches the contract from PRs #23, #29, #31, #34, #37, #43, #46, #71):
- Pre-exec try/except → returns `await orig_create(*args, **kwargs)` on telemetry setup failure
- `_safe()` helper swallows attribute-setting exceptions (per-resource basis)
- Real LLM errors still propagate after being recorded on the span
- AsyncStreamWrapper / SyncStreamWrapper for streaming with the same pattern

**Integration test** (`tests/integration/test_groq_instrumentation.py`): 330 lines, requires live `groq` SDK + `GROQ_API_KEY` to run. The test won't execute in this environment (no network), but it's there for CI.

Branch state: 16 commits behind `upstream/main @ 544deae`. Actual PR diff: 829 additions across `neatlogs/groq.py` (new file) + `neatlogs/instrumentation/registry.py` (1 line: register Groq) + `tests/integration/test_groq_instrumentation.py` (new file).

Harsh23Kashyap:
Follow-up: caught two consistency issues while reviewing the groq wrapper against the same pattern that surfaced the PR #71 finalize-leak bug.

**Issue 1: pre-exec fail-open used `orig_create(...)` directly instead of `_telemetry_fallback(orig_create, ...)`.**

Both forms work for the no-arg call, but the helper is the canonical entry point that all the other fail-open PRs (#23, #29, #31, #34, #37, #43, #46, #71) use. Groq should match.

**Issue 2: post-LLM-call finalize was a bare call with no try/except.**

If `_finalize_chat` threw, the exception propagated to the caller and the span was leaked (same shape as the PR #71 finalize-leak bug — pre-fix the openai/anthropic/bedrock/google_genai/openrouter/vertex_ai wrappers all had this; PR #71 fixed them for the per-call pre-exec fail-open contract but only fixed the leak on PR #71's own files). Groq was the only newer wrapper still vulnerable.

**Fix:**

- Imported `_safe_finalize` and `_telemetry_fallback` from `_wrap_utils`.
- Switched the pre-exec fallback to `_telemetry_fallback(orig_create, ...)`.
- Switched the finalize call to `_safe_finalize(span, _finalize_chat, response, duration_ms)`.
- Ported `_telemetry_fallback` and `_safe_finalize` to this branch's `_wrap_utils.py` (groq branch predates PR #71, so the helpers weren't here).

**New test:**

`test_groq_finalize_throws_still_ends_span_and_returns_response` forces `_finalize_chat` to throw and asserts:
1. response is returned to the caller
2. span is in the exporter (was leaked before)
3. span end_time is set
4. span status is OK

108 unit + 8 integration tests pass (was 7 integration). black + isort + ruff clean.

Branch: `feat/groq-wrapper @ d6aaa9f`. Pushed.

PR #61, opened 2026-07-30
feat(cost): add pricing {list,show} CLI subcommands

Harsh23Kashyap:
## What

Adds two subcommands to `neatlogs-cost` for catalog discovery without a span log:

```bash
neatlogs-cost pricing list [--format text|json|csv] [--pricing-file PATH]
neatlogs-cost pricing show <model_key> [--format text|json] [--pricing-file PATH]
```

Closes #60.

## Heads-up on base branch

This branch is based on `feat/cost` (the head of #59), but that branch only lives on the fork, so this PR had to be opened against `main`. As a result, the diff against `main` includes the cost module from #59 plus the four new commits here.

The view that isolates just this PR's work is `Harsh23Kashyap/feat/cost...Harsh23Kashyap/feat/cost-pricing-list-show` — four commits:
- `catalog()` method on the `PricingProvider` chain
- `format_pricing_list` and `format_pricing_show` formatters
- `pricing list` and `pricing show` CLI subcommands
- cleanup (dropped dead code)

After #59 merges to `main`, this PR's effective diff will shrink to those four commits. Reviewers can either look at the cross-fork compare now, or wait for #59 to land and then review the reduced diff.

## Example

```
$ neatlogs-cost pricing list
provider       model                                           context capabilities
-----------------------------------------------------------------------------------
anthropic      anthropic/claude-3-5-sonnet-latest              200,000 json_mode,prompt_cache,tools,vision
openai         openai/gpt-4o-mini                             128,000 json_mode,prompt_cache,tools,vision
...
33 model(s) listed.

$ neatlogs-cost pricing show anthropic/claude-sonnet-4
provider:        anthropic
context_window:  200,000
capabilities:    json_mode, prompt_cache, tools, vision

rates (USD per 1M tokens):
  cache_read     $ 0.300000
  cache_write    $ 3.750000
  input          $ 3.000000
  output         $15.000000

tiers:
  input          >    200,000 → $ 6.000000
  output         >    200,000 → $22.500000
```

## Changes

4 commits, ~800 LOC, 0 new dependencies, backward compatible.

- `neatlogs/cost/pricing.py` — new `catalog()` method on `PricingProvider` (default `[]`), overridden in `BuiltinProvider`, `CustomProvider`, and `ChainProvider` (deduped union, first match wins for keys present in multiple providers).
[...]

PR #59, opened 2026-07-26
feat(cost): add neatlogs-cost LLM cost intelligence engine

Harsh23Kashyap:
Closes #55.

## What

`neatlogs-cost` is a single CLI / module that answers the three cost questions a finance or engineering team actually asks:

1. **Per-model cost breakdown** — "what did this workload cost on each model, split by input / output / cache / reasoning?"
2. **What-if ranking** — "given the same workload, which candidate is the cheapest one that can still serve it?"
3. **Monthly forecast** — "what will my monthly bill be at a given call volume?"

All three share one schema and one CLI. Pricing is decoupled via a `PricingProvider` chain so users can override rates per-environment, fetch a remote catalog, or plug in LiteLLM's mirror without touching core code.

## Why

The original `neatlogs-compare` proposal in #55 had a flat `input_per_1m` / `output_per_1m` schema with boolean capability flags. That breaks down the moment a real user shows up:

- Pricing isn't flat anymore (`cache_read`, `cache_write`, `reasoning`, `image`, `audio`, long-context tiers).
- Capability flags don't compose (adding `supports_audio_input` is a schema + code change).
- Bundled pricing doesn't scale to 800+ models across 40 providers.
- Capability needs are per-workload, not per-user (the user is the one who knows whether they used vision, not the model they happened to call).
- Forecasting needs the same pricing data and the same tier handling; building it as a separate module would mean two schemas and two engines for one user question.

This PR ships one engine, one schema, and one CLI that handles all three.

## How

- **`PricingProvider` chain** — `CustomProvider` (user override) → `BuiltinProvider` (bundled `pricing.json`) → future providers. The chain pattern matches Langfuse's model-definition resolution: override wins, fall through to bundled for everything else, pluggable without touching the engine.
- **`usage_types` dict** — `input` / `output` / `cache_read` / `cache_write` / `reasoning` / `image` / `audio` / `embedding` / `...`. Open-ended; adding `per_image` or `gpu_seconds` is a JSON edit, not a code change.
- **`capabilities` as a set of strings** — `vision`, `tools`, `json_mode`, `prompt_cache`, `reasoning`, `audio`, `image_input`, `embedding`. Third-party providers can extend without code changes; the compatibility check fails closed on explicit declarations.
- **Capability auto-infer from source models** — a log that ran on `gpt-4o-mini` is assumed to need whatever `gpt-4o-mini` supports. `--need vision --need tools` overrides. `--no-auto-infer` turns it off entirely.
[...]

PR #62, opened 2026-07-30
fix(cost): three bug fixes from A-to-Z review

Harsh23Kashyap:
## What

Three bug fixes found by external review of the cost engine A-to-Z
knowledge doc. The engine shipped in #59 and the catalog discovery
landed in #61; this PR fixes three behavioral issues the prior reviews
missed.

## Heads-up on base branch

This branch is based on `feat/cost` (the head of #59), but that branch
only lives on the fork, so this PR is opened against `main` like #61.
The diff against `main` includes the cost work from #59 plus the
three fix commits here.

The view that isolates just this PR's work is
`Harsh23Kashyap/feat/cost...Harsh23Kashyap/fix/cost-bugs-from-a2z-review`.
Three commits. After #59 merges to `main`, this PR's effective
diff will shrink to those three commits.

### 1. Cost parser ignored the canonical `cache_write` key

`spans.py` read `neatlogs.llm.token_count.cache_creation` (and the
OTel raw `gen_ai.usage.cache_creation_input_tokens`), but the SDK's
canonical normalized attribute is `neatlogs.llm.token_count.cache_write`
(see `neatlogs/config/attribute-mapping.json`). Processed span logs
written by the SDK use the `cache_write` key. As shipped, the cost
engine undercounted cache_write cost as $0 for any real workload.

**Fix.** Read `cache_write` first; fall back to `cache_creation` /
`cache_creation_input_tokens` only for legacy / raw OTel records.

**Rename.** `SpanUsage.cache_creation_tokens` → `cache_write_tokens` to
match the canonical key. The cost module is v1 with no merged users,
so the rename is clean. All internal references in `breakdown.py` and
`workload.py` updated; tests updated.

### 2. `--min-context` was an on/off flag, not a threshold

`ranking.py:_is_span_compatible` only checked
`model.context_window < usage.input_total`. The configured
`min_context_window` value was unused. With `--min-context 200000`
and a 128K-context model, the model was marked compatible as long
as no individual prompt exceeded 128K.

**Fix.** Added a check between the existing branches:
`elif model.context_window < constraints.min_context_window`.

### 3. Forecast cache fallback was silent

When a model declared `cache_write` but no `cache_read` rate, the
forecast fell back to `cache_write` for cache-hit traffic without
saying so. A user could see a non-zero cache cost and not realize
their model has no cache_read rate.

**Fix.** Append a note to the forecast report when the fallback is
used. Behavior (compute the cost) is unchanged for backward
compatibility; only visibility changes. A future v2 may add an
explicit `cache_write_per_call` input to model this separately.

## Tests

| Test | Purpose |
|---|---|
| `test_cache_write_canonical_key` | Span with `neatlogs.llm.token_count.cache_write` extracts to `cache_write_tokens` correctly. |
| `test_cache_write_key_takes_precedence_over_legacy` | When both `cache_write` and `cache_creation` are present, `cache_write` wins. |
| `test_min_context_floor_enforced_independent_of_prompt_size` | 128K-context model with 1K prompt and `min=200K` is incompatible. |
| `test_cache_read_falls_back_to_cache_write_with_note` | Model with `cache_write` but no `cache_read` produces a cache cost and a note mentioning both keys. |

## Verification

- 145 cost tests pass.
- `black` and `isort` clean.
- 96% coverage on `neatlogs/cost`.
- 0 new dependencies.
- Backward compatible: the `SpanUsage` field rename is the only
  public surface change; the cost module is v1 with no merged users.

## Out of scope

The cache_write semantic gap (a `LiteLLMProvider` and a parity test
to verify the bundled catalog) are in the forward-work table of the
cost engine wiki; both are out of scope here.


tanisha-neatlogs:
@Harsh23Kashyap , The existing instrumentation already emits `cache_write` correctly, closing this.

Harsh23Kashyap:
Hey Tanisha, makes sense, the existing instrumentation already emits `cache_write` correctly. Closing along with the others.

The parity test against LiteLLM (with the weekly CI cron) is the main reusable artifact if you want to bolt it onto the internal catalog as a drift check.

PR #66, opened 2026-08-06
fix(crewai): preserve tool telemetry fidelity

khushalkumar:
## Summary

Preserve complete CrewAI tool inputs and outputs for backend payload handling, and add explicit CrewAI framework identity to SDK-created spans.

## Problem

The BaseTool and CrewStructuredTool wrappers sliced tool input and output at 10,000 characters before export. Long structured values became incomplete fragments, so the backend never received enough data to apply its payload and S3 policy. CrewAI wrapper spans also omitted `neatlogs.framework=crewai`.

## Changes

- serialize CrewAI tool I/O as full JSON with `default=str` and no 10k or shared 100k cutoff;
- preserve original return values and exception behavior even when telemetry serialization fails;
- use the same capture policy for BaseTool and CrewStructuredTool;
- add `neatlogs.framework=crewai` to Crew, Flow, Task, Agent, Tool, and LLM wrapper spans;
- add regression coverage for input/output beyond 100k, structured-tool parity, empty results, datetime fallback, serialization failure, errors, and representative framework markers.

## Validation

- `pytest -q tests/unit`: 98 passed
- `pytest -q tests/unit/test_crewai_tool_fidelity.py`: 9 passed
- `black --check neatlogs tests`: passed
- `isort --check-only neatlogs tests`: passed
- `git diff --check`: passed
- CrewAI integration suite: 19 passed, 1 failed because locked CrewAI 1.5.0 does not expose `Crew.akickoff`; the failing test is unchanged by this PR.

## Cross-repository acceptance required before release

This PR is intentionally a draft. Adding the explicit CrewAI marker may activate the backend framework-specific finalizer path. Before merge/release, validate a representative WORKFLOW -> AGENT -> LLM -> TOOL trace against the backend finalizer and confirm both full tool output and the complete user task remain visible. The backend repository is not among the repositories visible to the authenticated `khushalkumar` GitHub account, so that acceptance could not be run here.

## Additional audit findings outside this PR

- 14 explicit 10k slices remain across Crew, Task, Agent, Flow, and LLM paths.
- Consolidated CrewAI LLM `input.value` still uses the shared 100k serializer; a 120k reproduction lost its tail and produced invalid JSON.
- Sync batch kickoff records `output.value`, while async batch kickoff does not.
- Crew/Flow input and Crew/Task output truthiness checks can omit or replace explicit empty values.
- The declared CrewAI and LiteLLM lower bounds resolve to an incompatible OpenAI dependency pair; the lockfile combination works, apart from the existing `akickoff` test mismatch.

## Release impact

- no schema migration;
- no version bump;
- no package publication;
- no change to tool business behavior;
- historical truncated traces cannot recover missing tails.

### DIFF
--- a/neatlogs/crewai.py
+++ b/neatlogs/crewai.py
@@ -45,6 +45,24 @@ _CLASS_HOOKS_INSTALLED = False
 _LLM_INPUT_CAPTURE_LIMIT = 1 * 1024 * 1024
 
 
+def _crewai_span_attributes(kind: str) -> dict[str, Any]:
+    return {
+        "neatlogs.span.kind": kind,
+        "neatlogs.framework": "crewai",
+    }
+
+
+def _serialize_crewai_io(value: Any) -> str:
+    """Serialize CrewAI tool I/O without discarding data before export."""
+    try:
+        return json.dumps(value, default=str, ensure_ascii=False)
+    except Exception:
+        try:
+            return str(value)
+        except Exception:
+            return f"<unserializable {type(value).__name__}>"
+
+
 def _shared_trace_kwargs() -> dict:
     """Context for the crewai.crew.kickoff root span.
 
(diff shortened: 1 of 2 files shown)


khushalkumar:
Superseded by #67 to correct the commit attribution to khushalkumar and remove the unintended codex/ branch prefix.

PR #69, opened 2026-08-13
feat: neatlogs.annotate() and neatlogs.add_event() for post-hoc trace enrichment

### LINKED ISSUE #68: Add neatlogs.annotate() and neatlogs.add_event() for post-hoc trace enrichment
Title: Add neatlogs.annotate() and neatlogs.add_event() for post-hoc trace enrichment

Right now there's no way to attach metadata to a span after it's recorded. You have to know what to attach at span time, which doesn't work for:

- user feedback coming in after the fact (LLM produced a bad response, team wants to mark the trace needs_review=true, severity=high, ticket=JIRA-1234)
- A/B test cohort, recorded after the response
- automated evaluator quality_score, attached after the LLM output was judged

Proposal: two new top-level functions on the neatlogs module.

    import neatlogs

    # filterable attrs under neatlogs.annotation.*
    applied = neatlogs.annotate(needs_review=True, severity="high", ticket="JIRA-1234")

    # time-stamped event marker
    applied = neatlogs.add_event("user_feedback", rating=1, comment="wrong answer")

Both return True if at least one was applied, False otherwise. Both silently no-op (return False, no raise) when init() wasn't called, no active neatlogs span, the span is foreign (co-tenant), or the span has ended / is sampled out.

Namespacing: all keys and event names get the neatlogs.annotation. prefix so they're filterable in the backend without colliding with what the wrappers emit. Example on a span: neatlogs.annotation.needs_review=true, plus an event neatlogs.annotation.user_feedback with rating=1, comment="wrong answer".

Value coercion:
- primitives (str, int, float, bool) pass through native so backend numeric/boolean filters stay meaningful
- complex objects (dict, list, datetime, Pydantic v2) get JSON-serialized
- NaN/Inf get stringified so strict parsers (ClickHouse) don't reject the attr

Safety: resolver reuses the existing _isolation_active, _current_neatlogs_parent, _is_neatlogs_span from neatlogs/_wrap_utils.py. The naive get_current_span would leak into a foreign span when a co-tenant pipeline (openlit, langfuse) is running alongside — same isolation-aware pattern as the 7 fail-open wrapper PRs (#23, #29, #31, #34, #37, #43, #46).

Scope:
- new file neatlogs/_annotations.py (~210 lines)
- new file tests/unit/test_annotations.py (59 unit cases)
- edit neatlogs/__init__.py (2 exports + __all__ entry)
- no changes to existing files
- no new dependencies

Out of scope for v1:
- wiring into @span (would change root-eligibility logic, too invasive)
- with neatlogs.annotating(...) context manager (zero capability gain)
- per-trace_id post-export annotation (would need a backend write endpoint)

Open question: do you want only a specific well-known schema (e.g. ticket, severity, needs_review) to be indexable, or any neatlogs.annotation.* key fine? Current PR emits all keys. If you want a constrained schema, can add it as a follow-up once the schema is decided.

Harsh23Kashyap:
Hey Tanisha, new feature for post-hoc trace enrichment. Two new top-level functions on the neatlogs module:

    import neatlogs

    # filterable attrs under neatlogs.annotation.*
    applied = neatlogs.annotate(needs_review=True, severity="high", ticket="JIRA-1234")

    # time-stamped event marker
    applied = neatlogs.add_event("user_feedback", rating=1, comment="wrong answer")

Both return True if at least one was applied, False otherwise. Both silently no-op (return False, no raise) when init() wasn't called, no active neatlogs span, the span is foreign (co-tenant), or the span has ended / is sampled out.

Why: production debugging workflows that need to mark a trace after the fact — user feedback came in late, evaluator scored it poorly, etc. — without re-running the LLM call. The neatlogs.annotation.* namespace keeps annotations filterable in the backend without colliding with wrapper-emitted attrs.

Design decisions:

isolation-aware span resolver. Reuses the existing _isolation_active, _current_neatlogs_parent, _is_neatlogs_span from neatlogs/_wrap_utils.py. The naive get_current_span would leak into a foreign span when a co-tenant pipeline (openlit, langfuse) is running alongside — the new code explicitly refuses to do that.

bool return, not raise. Fail-open. Annotation should never crash user code. The bool return makes the no-op testable.

primitives pass through native. int stays int so backend numeric/boolean filters stay meaningful. Complex objects (dict, list, datetime, Pydantic) get JSON-serialized. NaN/Inf get stringified so strict JSON parsers (ClickHouse) don't reject the attr.

key validation. Dotted keys rejected because the backend's ClickHouse mapping splits them into nested structures. Keys starting with neatlogs. rejected to prevent double-namespacing. Empty / non-string keys rejected.

no wiring into @span. Would change root-eligibility logic (_ROOT_KINDS), simplified-view mapping, and backend rooting. Too invasive for v1.

Tests: 59 unit cases in tests/unit/test_annotations.py — active span, no span, no init, foreign span (default and isolated), ended span, dotted keys, empty keys, reserved prefix, None values, primitive type preservation, complex type JSON coercion, NaN/Inf, empty kwargs, partial-application, invalid event names, Pydantic model_dump, thread-context propagation, repeated-call accumulation, asyncio task, multi-thread (8 workers, context-captured and re-attached), @span decorator integration, identify() context, log() interleaving, public API surface (dir, __all__), 1000-attr stress, 1000-event stress, no-double-namespace, chain semantics. 167/167 unit tests pass on the branch. black + isort clean.

One question for you on the backend: PR emits arbitrary neatlogs.annotation.* keys. The existing neatlogs.* mapping in attribute_processor.py:917 should pick them up at the OTel layer, but filterability is a backend-indexing question not an OTel question. If you want only a specific well-known schema (ticket, severity, needs_review) to be indexable while keeping free-form keys filterable-but-unindexed, can add a constrained schema as a follow-up once the schema is decided. Fine either way.

Closes #68. Branch feat/trace-annotations is on the fork, ready for review.

### DIFF
--- a/neatlogs/__init__.py
+++ b/neatlogs/__init__.py
@@ -26,6 +26,7 @@ Available span kinds:
     - "MCP_TOOL" - MCP protocol tool (auto Pydantic handling)
 """
 
+from ._annotations import add_event, annotate
 from .client import Client
 from .core.context import trace
 from .core.crewai_task_registry import register_crewai_task
@@ -280,6 +281,8 @@ __all__ = [
     "shutdown",
     "span",
     "trace",
+    "annotate",
+    "add_event",
     "identify",
     "inject_trace_context",
     "extract_trace_context",
--- a/neatlogs/_annotations.py
+++ b/neatlogs/_annotations.py
@@ -1,0 +1,219 @@
+"""
+Post-hoc trace annotations.
+
+Public API:
+    import neatlogs
+
+    applied = neatlogs.annotate(needs_review=True, severity="high", ticket="JIRA-1234")
+    applied = neatlogs.add_event("user_feedback", rating=1, comment="wrong answer")
+
+Both functions silently no-op (return False) when:
+- neatlogs.init() was not called
+- there is no active neatlogs span
+- the active span is a foreign (co-tenant) span — prevents annotation leak
+- the active span has ended or is sampled out
+- the key/name is invalid (dotted, starts with "neatlogs.", empty, non-string)
+
+Both functions return True if at least one attribute/event was applied.
+
+All attributes and event names are namespaced under "neatlogs.annotation." so
+they can be filtered in the backend without colliding with instrumentation
+attributes emitted by the provider wrappers.
+"""
+
+from __future__ import annotations
+
+import json
+import math
+from typing import Any, Dict, Optional
+
+from opentelemetry import trace as otel_trace
+from opentelemetry.trace import Span
+
+from ._wrap_utils import (
+    _current_neatlogs_parent,
+    _is_neatlogs_span,
+    _isolation_active,
+)
+
+ANNOTATION_PREFIX = "neatlogs.annotation."
+
+
+# Valid key characters: ASCII letters, digits, underscore, hyphen. Mirrors the
+# loosest convention OpenTelemetry itself allows for attribute keys; explicitly
+# excludes dots (which the backend's ClickHouse mapping would split into nested
+# structures) and the "neatlogs." prefix (which would cause double-namespacing).
+_VALID_KEY_CHARS = set("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-")
+
+
+def _valid_key(key: Any) -> bool:
+    if not isinstance(key, str) or not key:
+        return False
+    if "." in key:
+        return False
+    if key.startswith("neatlogs."):
+        return False
+    return all(c in _VALID_KEY_CHARS for c in key)
+
+
+def _coerce(value: Any) -> Any:
+    """Pass OTel-native primitives through; JSON-serialize complex objects.
+
+    bool is checked before int because bool is a subclass of int in Python.
+    NaN/Inf are converted to their string form because json.dumps emits them
+    by default and strict backend parsers (e.g. ClickHouse JSON) reject the
+    unquoted forms. Pydantic v2 models are dumped via .model_dump() first.
+    """
+    if value is None:
+        return None
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, (int, str)):
+        return value
+    if isinstance(value, float):
+        if math.isnan(value) or math.isinf(value):
+            return "NaN" if math.isnan(value) else ("Infinity" if value > 0 else "-Infinity")
+        return value
+    if isinstance(value, bytes):
+        # OTel supports bytes-as-string; pass through and let the SDK decide
+        # whether to base64-encode. utf-8 decode with errors='replace' as a
+        # safety net; fall back to repr for non-decodable bytes.
+        try:
+            return value.decode("utf-8")
+        except (UnicodeDecodeError, AttributeError):
+            return repr(value)
+    if isinstance(value, (dict, list, tuple)):
+        return json.dumps(value, default=_pydantic_default)
+    # datetime, Pydantic v2 models, or anything else: try .model_dump() first.
+    if hasattr(value, "model_dump"):
+        try:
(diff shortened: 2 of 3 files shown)


Harsh23Kashyap:
Verified locally on `feat/trace-annotations @ c277edb` (3 commits):

**Test results:** 178 unit tests pass, 2 xfailed (expected, documented in test_annotations.py). Linters clean on the 2 changed files (3 unused-imports + 1 f-string issue found and fixed in `c277edb`).

**Design verification:**

Public API (matches docstring):
- `neatlogs.annotate(**attrs) -> bool` — applies attrs as `neatlogs.annotation.<key>` on the current span
- `neatlogs.add_event(name, **attrs) -> bool` — adds an event with `neatlogs.annotation.<name>` and the attrs

Both silently return `False` when:
- `neatlogs.init()` was not called
- no active neatlogs span (foreign or missing)
- the active span is a foreign (co-tenant) span (prevents annotation leak)
- the active span has ended or is sampled out
- the key/name is invalid (dotted, starts with `neatlogs.`, empty, non-string)

Both return `True` if at least one attribute/event was applied.

**The fail-open contract** (the second commit, 567d9ae) is honored:
- Coercion failure on a value (e.g. circular ref, hostile `__str__`) → skip just that value
- `set_attribute` / `add_event` SDK-level failure → skip just that value
- One bad key never breaks the others

**Type coercion** mirrors `_coerce_workflow_value` from PR #70:
- primitives pass through natively
- NaN/Inf are stringified
- bytes → utf-8 decode (with `errors='replace'` fallback to `repr()`)
- dict/list/tuple → JSON-serialized
- Pydantic v2 models → `model_dump()` first
- datetimes → `isoformat()`

Branch state: 20 commits behind `upstream/main @ 544deae`. Actual PR diff: 1642 additions / 0 deletions across `neatlogs/_annotations.py` (new) + `tests/unit/test_annotations.py` (358→1063 lines) + `neatlogs/__init__.py` (3 new exports).

Harsh23Kashyap:
Follow-up: fixed both F2 and F5 from the consult-kiro panel findings. They were tracked as xfail with `strict=True` in `test_annotations.py`, meaning the bugs were documented but the test suite flagged them. Both bugs are real.

**F2: nested NaN/Inf not stringified.**

The pre-existing `_coerce` handled top-level NaN/Inf floats by stringifying them, but didn't recurse into dict/list/tuple values. So `annotate(payload={"score": math.nan, "limit": math.inf})` produced a JSON string with bare `NaN` / `Infinity` tokens (json.dumps allow_nan=True), which strict JSON parsers (e.g. ClickHouse) reject. The top-level float case worked, the nested case didn't.

Fix: added `_sanitize_nan_inf()` (same pattern as PR #70's helper) and routed dict/list/tuple through it before `json.dumps`. Also applied it to the Pydantic v2 `model_dump()` path. Non-float values pass through; tuples are returned as tuples so the JSON encoder still produces an array.

**F5: oversized values not truncated.**

The rest of the SDK truncates values to 100KB via `serialize()` in `_wrap_utils.py` to keep spans bounded. `_coerce` skipped that step, so a 200KB string landed whole on the span. The xfail test asserted `len(raw) < 110_000` and failed.

Fix: imported `serialize` and routed long strings (len > 100_000) through it in `_coerce`. Native short strings still pass through unchanged (no JSON round-trip), preserving the current "string in, string out" contract for the common case.

**Test impact:**

- 2 xfail markers removed from `test_annotations.py`; tests now run as plain asserts.
- 1 unused `import pytest` removed from the test file.
- 72/72 tests pass (was 70 + 2 xfailed).
- The "consult-kiro panel findings" comment block updated to reflect that F2/F5 are fixed.

Branch: `feat/trace-annotations @ b4ecbb3`. Pushed.

PR #70, opened 2026-08-14
fix: preserve native types in neatlogs.workflow.* attributes

Harsh23Kashyap:
Bug found while looking through the wrap() flow. `apply_wrap_context_attributes` in `neatlogs/_wrap_utils.py:266` was running `str(value)` on every workflow attribute before stamping it on the auto-created WORKFLOW root span. That silently corrupted non-string values:

- `count=42` (int) was stored as the string `"42"`
- `enabled=True` (bool) was stored as the string `"True"`
- `tags=["a", "b"]` (list) was stored as `"['a', 'b']"` (Python repr, not JSON)
- `ratio=0.95` (float) was stored as the string `"0.95"`

The public API is `neatlogs.wrap(client, **workflow_attributes)`, and the docstring says workflow kwargs land on the root span as `neatlogs.workflow.*` "so they show in the metadata drawer and trace search." "Trace search" needs filterable types — numeric, boolean, list-shaped filters on the stored attributes all fail today.

Why this slipped: no test exercised the wrap-context path with non-string values. Grepped tests/ to confirm.

Inconsistency: the sibling function `_set_span_attributes` in `neatlogs/core/context.py:281` does not stringify. So `neatlogs.trace("x", count=42)` works correctly (stores int 42), but `neatlogs.wrap(client, count=42)` does not. Same intent, two different code paths, two different behaviors.

Reproducer (before fix):

    import neatlogs
    from neatlogs._wrap_utils import _wrap_context, apply_wrap_context_attributes

    ctx_token = _wrap_context.set({"workflow": {
        "count": 42, "enabled": True, "tags": ["a", "b"], "ratio": 0.95,
    }})
    # ... apply to a span via the auto-root path ...

    # BEFORE FIX:
    #   neatlogs.workflow.count   = '42'    (str)
    #   neatlogs.workflow.enabled = 'True'  (str)
    #   neatlogs.workflow.tags    = "['a', 'b']"   (Python repr)
    #   neatlogs.workflow.ratio   = '0.95'  (str)

After fix (this commit):

    # AFTER FIX:
    #   neatlogs.workflow.count   = 42     (int)
    #   neatlogs.workflow.enabled = True   (bool)
    #   neatlogs.workflow.tags    = '["a", "b"]'  (JSON)
    #   neatlogs.workflow.ratio   = 0.95   (float)

Fix: replace `str(value)` with a new `_coerce_workflow_value` helper. Primitives (bool, int, str, float) pass through natively so backend numeric/boolean filters work. `NaN`/`Inf` floats get stringified so strict JSON parsers (ClickHouse) don't reject the attribute. `dict` / `list` / `tuple` get JSON-serialized. `bytes` are utf-8 decoded. Everything else falls back to `str()`. Mirrors `neatlogs._annotations._coerce` so `annotate()` and `wrap()` agree on type semantics. The helper is duplicated rather than imported because `_annotations` already depends on `_wrap_utils` (circular import).

Tests: 10 new cases in `tests/unit/test_wrap_context_attributes.py` cover int / bool / float / list / dict / NaN+Inf / fail-open on OTel errors / is_root gating / empty context. All 118 unit tests pass on this branch. `black` + `isort` clean.

Out of scope for this fix (separate concerns, can be follow-ups if wanted):
- `_set_span_attributes` in `core/context.py:281` also passes complex values to `span.set_attribute()` without coercion — OTel will either accept or silently drop them. The `try/except` is not present there, so a bad value would propagate. Different bug, different file.
- The annotations feature (PR #69) has its own `_coerce` that already handles this correctly.

Happy to split the test file or amend the fix if you want a different shape. Direct, no questions.

### DIFF
--- a/neatlogs/_wrap_utils.py
+++ b/neatlogs/_wrap_utils.py
@@ -280,9 +280,14 @@ def _coerce_workflow_value(value: Any) -> Any:
     - Primitives (bool, int, str, float) pass through natively so backend
       numeric / boolean filters remain meaningful.
     - ``NaN`` / ``Inf`` floats are stringified so strict JSON parsers
-      (e.g. ClickHouse) do not reject the attribute.
+      (e.g. ClickHouse) do not reject the attribute. This applies
+      recursively inside dict / list / tuple as well.
     - Bytes are utf-8 decoded; unrepresentable bytes fall back to repr.
     - dict / list / tuple are JSON-serialized.
+    - Pydantic-v2-style objects (anything with ``model_dump()``) are
+      dumped then JSON-serialized.
+    - datetimes (anything with ``isoformat()``) are stringified via
+      ``isoformat()``.
     - Everything else falls back to ``str()``.
     """
     if isinstance(value, bool):
@@ -300,10 +305,44 @@ def _coerce_workflow_value(value: Any) -> Any:
             return repr(value)
     if isinstance(value, (dict, list, tuple)):
         try:
-            return json.dumps(value, default=str)
+            return json.dumps(_sanitize_nan_inf(value), default=str)
         except (TypeError, ValueError):
             return str(value)
+    if hasattr(value, "model_dump"):
+        try:
+            return json.dumps(_sanitize_nan_inf(value.model_dump()), default=str)
+        except Exception:
+            pass
+    if hasattr(value, "isoformat"):
+        try:
+            return value.isoformat()
+        except Exception:
+            pass
     return str(value)
+
+
+def _sanitize_nan_inf(obj: Any) -> Any:
+    """Recursively replace NaN / Inf floats with their string forms so
+    ``json.dumps(allow_nan=True)`` does not emit bare ``NaN`` / ``Infinity``
+    tokens that strict JSON parsers (e.g. ClickHouse) reject.
+
+    Pass-through for every other type. Tuples are returned as tuples so the
+    outer JSON encoder still produces an array; dicts and lists are
+    rebuilt so the recursion actually walks nested structures.
+    """
+    if isinstance(obj, float):
+        if math.isnan(obj):
+            return "NaN"
+        if math.isinf(obj):
+            return "Infinity" if obj > 0 else "-Infinity"
+        return obj
+    if isinstance(obj, dict):
+        return {k: _sanitize_nan_inf(v) for k, v in obj.items()}
+    if isinstance(obj, list):
+        return [_sanitize_nan_inf(v) for v in obj]
+    if isinstance(obj, tuple):
+        return tuple(_sanitize_nan_inf(v) for v in obj)
+    return obj
 
 
 def reset_tracer() -> None:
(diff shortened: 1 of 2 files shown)


Harsh23Kashyap:
Verified locally on `fix/wrap-context-attributes-coercion @ 701a1bf` (2 commits):

**Test results:** 129 unit tests pass on Python 3.11. Linters (black + isort + ruff) clean on the 1 changed file.

**Behavior verification** (synthetic, no network needed):

- `bool` / `int` / `str` / `float` pass through natively
- `NaN` / `Inf` / `-Inf` floats → stringified to `"NaN"` / `"Infinity"` / `"-Infinity"`
- `bytes` → utf-8 decode, with `repr()` fallback for invalid sequences
- `dict` / `list` / `tuple` → JSON-serialized
- Nested `NaN`/`Inf` inside `dict`/`list` recursively stringified (via `_sanitize_nan_inf`)
- `datetime` → `isoformat()` (e.g. `"2026-01-15T12:30:00"`)
- Pydantic-v2-style objects (anything with `.model_dump()`) → dumped then JSON-serialized
- Unknown class → `str()`

The design correctly preserves native types (so backend numeric/boolean filters remain meaningful) while handling the strict-JSON-parser edge cases (ClickHouse rejects bare `NaN`/`Infinity` tokens).

**Diff scope:** 1 file (`neatlogs/_wrap_utils.py`) — adds `_coerce_workflow_value` + `_sanitize_nan_inf`. Mirrors `neatlogs._annotations._coerce` but duplicated to avoid a circular import (the `_annotations` module already depends on `_wrap_utils`).

Branch state: 21 commits behind `upstream/main @ 544deae` (uv migration + OTel 1.35.0 floor). Actual PR diff: 353 additions / 1 deletion across 1 file.

Harsh23Kashyap:
Follow-up: caught a docstring inaccuracy while reading the diff.

The docstring on `_coerce_workflow_value` claimed it "Mirrors `neatlogs._annotations._coerce`" so the two helpers agree. They don't, and the divergence is intentional.

**Divergence on `None`:**

- `_annotations._coerce(None)` returns `None` (and the public `annotate()` filters it).
- `_coerce_workflow_value(None)` returns `"None"` via `str()`.

The reason: the wrap-context path can receive a `None` value from user-supplied workflow metadata, and OTel's `set_attribute` rejects `None`. Stringifying keeps the call from raising inside the outer `try/except: pass`. The existing test `test_wrap_context_skips_none_values` documents the choice and notes that the upstream `_filtered_mapping` is the real defense against `None` in the public API.

Fixed the docstring to:
1. State explicitly that this is a standalone copy, not a mirror.
2. Explain the `None` divergence and why it exists.
3. Tighten the NaN/Inf bullet (only float NaN/Inf are replaced; non-float values pass through).

No behavior change. 21/21 tests pass; black clean.

Branch: `fix/wrap-context-attributes-coercion @ 0ddafae`. Pushed.

PR #71, opened 2026-08-17
fix(per-call): fail open when wrapper pre-execution or finalize throws

### LINKED ISSUE #15: Telemetry errors can crash the host application due to lack of try-catch wrappers around hook pre-execution
### Description
While `BaseEventHandler` correctly wraps extraction functions in try-except clauses during enrichment, the pre-execution phase (like `self.handle_call_start()`, parameter extraction, and span initialization) lacks catch-all exception wrappers. 

If `self.handle_call_start` or any of its internal logic throws an unexpected exception (e.g., because a provider updates their SDK structure and a parameter type changes), the exception will propagate to the caller **before** the `original_method` is ever executed, completely crashing the user's primary application.

```python
# neatlogs/event_handlers/base.py (lines 168-172)
self.handle_call_start(span, *args, **kwargs) # <--- If this throws, original_method is never run!
try:
    response = original_method(*args, **kwargs)
    self.handle_call_end(span, response, success=True)
    return response
```

### Impact
- Critical failure of the host application. A telemetry or tracking library should **never** cause the primary business logic/LLM call to fail.

### Suggestion 
Wrap all initialization, hooking, and telemetry-start actions in a safety try-except block, logging any internal library errors as debug warnings while always falling back to executing and returning the `original_method`.

Harsh23Kashyap:
Fixes #15.

## Problem

Provider wrappers (OpenAI, Anthropic, Google GenAI, Vertex AI, Bedrock, OpenRouter) ran their pre-execution phase (`tracer.start_span`, input message iteration, tool schema serialization, `_set_input_attributes`, request metadata) outside the try/except that wrapped the original `create` / `generate_content` / `converse` / `invoke_model` / `send` call. If any of those internal calls raised, the exception escaped to the user, the host application's LLM call never ran, and a telemetry library crashed the primary business logic.

A telemetry library should never crash the host application.

Reproducer (before the fix):

```python
class ThrowingMsg:
    def get(self, key, default=None):
        raise RuntimeError("malformed message")
    def __getitem__(self, key):
        raise RuntimeError("malformed message")

client.chat.completions.create(
    model="gpt-4", messages=[ThrowingMsg()]
)
# RuntimeError raised by neatlogs, before orig_create is called.
# The user's LLM call never runs.
```

## Fix

Third leg of the fail-open contract:

1. wrap-time fail-open (PRs #23, #29, #31, #34, #37, #43, #46): telemetry SETUP falls back to bare SDK on failure
2. wrap-context fail-open (PR #70): `neatlogs.wrap(..., project_id=42)` does not stringify int
3. per-call fail-open (this PR): telemetry EXECUTION (pre-exec + finalize) falls back to bare SDK on failure

What changed:

- New helper `_telemetry_fallback(orig, *args, **kwargs)` in `neatlogs/_wrap_utils.py`.
- Every provider wrapper's pre-execution phase is now wrapped in `try/except`. On failure we return the bare orig result.
- The existing `try: orig_*(*a, **kw) / except / raise` block is preserved. Real LLM errors still propagate after being recorded on the span.
- The finalize phase (`_finalize_response`, `_finalize_chat`, etc.) is also wrapped. If it throws after we already have a response, the response still returns to the user.

No public API change, no behavior change on the success path.

## Affected wrappers

`neatlogs/openai.py`, `anthropic.py`, `google_genai.py`, `vertex_ai.py`, `bedrock.py`, `openrouter.py`. Every `patched_create` / `patched_generate_content` / `patched_converse` / `patched_invoke_model` / `patched_send` plus their async variants, plus the secondary methods (`parse`, `count_tokens`, `embed_content`, `send_message`, `send_message_stream`, `rerank`).

## Tests

`tests/unit/test_per_call_fail_open.py`, 11 new tests:

- `_telemetry_fallback` itself (calls orig, propagates orig errors)
- one wrapper per provider: pre-exec throws, orig still runs
- openai: real LLM exception from orig is still recorded on the span and re-raised (not swallowed)
- openai: happy path still emits the LLM span

Total unit suite on this branch: 119 passed (was 108 on the merge base, +11).

## Commits

- `fix(per-call): fail open when wrapper pre-execution or finalize throws`
- `test(per-call): add regression coverage for fail-open on pre-exec errors`

### DIFF
--- a/neatlogs/_wrap_utils.py
+++ b/neatlogs/_wrap_utils.py
@@ -589,6 +589,17 @@ def is_suppressed() -> bool:
         return False
 
 
+def _telemetry_fallback(orig: Callable[..., Any], *args: Any, **kwargs: Any) -> Any:
+    """Call the untraced original directly.
+
+    Used by provider wrappers when their own pre-execution or finalize phase
+    raises an exception. The user's primary LLM call must never be blocked
+    by a neatlogs internal bug; a telemetry library should fail open, not
+    crash the host application.
+    """
+    return orig(*args, **kwargs)
+
+
 # ---------------------------------------------------------------------------
 # Auto-root
 #
--- a/neatlogs/_wrap_utils.py
+++ b/neatlogs/_wrap_utils.py
@@ -600,6 +600,24 @@ def _telemetry_fallback(orig: Callable[..., Any], *args: Any, **kwargs: Any) -> 
     return orig(*args, **kwargs)
 
 
+def _safe_finalize(span: Any, finalize_fn: Callable[..., Any], *args: Any) -> None:
+    """Run a finalize callback; if it throws, end the span with OK so it is
+    not leaked, then swallow. Used by the per-call wrappers when the LLM
+    call has already returned a response and the only thing left is to
+    record telemetry attributes.
+    """
+    try:
+        finalize_fn(span, *args)
+    except Exception:
+        from opentelemetry.trace import StatusCode
+
+        try:
+            span.set_status(StatusCode.OK)
+            span.end()
+        except Exception:
+            pass
+
+
 # ---------------------------------------------------------------------------
 # Auto-root
 #
--- a/neatlogs/anthropic.py
+++ b/neatlogs/anthropic.py
@@ -22,6 +22,7 @@ from opentelemetry.trace import StatusCode
 from ._wrap_utils import (
     AsyncStreamWrapper,
     SyncStreamWrapper,
+    _safe_finalize,
     _telemetry_fallback,
     get_provider_tracer,
     is_suppressed,
@@ -117,10 +118,7 @@ def _patch_messages(messages: Any) -> None:
             return SyncStreamWrapper(response, span, _finalize_stream)
 
         duration_ms = (time.perf_counter() - start) * 1000
-        try:
-            _finalize_response(span, response, duration_ms)
-        except Exception:
-            pass
+        _safe_finalize(span, _finalize_response, response, duration_ms)
         return response
 
     messages.create = patched_create
@@ -207,10 +205,7 @@ def _patch_async_messages(messages: Any) -> None:
             return AsyncStreamWrapper(response, span, _finalize_stream)
 
         duration_ms = (time.perf_counter() - start) * 1000
-        try:
-            _finalize_response(span, response, duration_ms)
-        except Exception:
-            pass
+        _safe_finalize(span, _finalize_response, response, duration_ms)
         return response
 
     messages.create = patched_create
@@ -604,10 +599,7 @@ def _extra_message_methods(messages: Any, is_async: bool) -> None:
                 except Exception as e:
                     _err(span, e)
                     raise
-                try:
-                    _finalize_response(span, resp, (time.perf_counter() - start) * 1000)
-                except Exception:
-                    pass
+                _safe_finalize(span, _finalize_response, resp, (time.perf_counter() - start) * 1000)
                 return resp
 
         else:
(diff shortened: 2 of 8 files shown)


Harsh23Kashyap:
Verified locally on `fix/per-call-pre-exec-fail-open @ feebb71` (2 commits):

**Test results:** 119 unit tests pass, including the 11 new `test_per_call_fail_open.py` cases (pre-exec throwing for OpenAI, Anthropic, Google GenAI, Vertex AI, Bedrock, OpenRouter — all 6 provider wrappers).

**Linters:** black + isort + ruff clean on the 7 changed files.

**Behavior verification:**

Pre-exec fail-open (the new helper):
- `_telemetry_fallback(orig, *args, **kwargs)` calls the untraced original
- If `orig` is None or non-callable, the helper would raise — but `orig` is always the untraced provider function, so this is not a real concern in practice
- 11 unit tests verify the helper for all 6 providers

Finalize fail-open (pre-existing pattern, unchanged by this PR):
- The finalize phase (`_finalize_response`, `_finalize_responses_response`, etc.) is already wrapped in `try/except: pass` in the wrapper functions — a pre-existing pattern, not added by this PR
- The PR body claim is correct: "if it throws after we already have a response, the response still returns to the user"
- Synthetic test confirms: simulating `_finalize_response` throwing inside the wrapper's try/except doesn't crash the host app

The MMX review raised a question about whether the finalize phase was wrapped, but the code does wrap it (just with a different pattern — `try/except: pass` rather than `_telemetry_fallback`, which is correct since the response is already in hand).

The PR's actual changes:
- 1 new helper in `_wrap_utils.py` (`_telemetry_fallback`)
- 6 wrappers wrapped in pre-exec try/except (anthropic, bedrock, google_genai, openai, openrouter, vertex_ai)
- 11 unit tests in `test_per_call_fail_open.py`
- Real LLM errors still propagate after being recorded on the span (the original `try: orig / except / raise` is preserved)

Branch state: 22 commits behind `upstream/main @ 544deae` (uv migration + OTel 1.35.0 floor). The 11970-line diff against upstream is noise from those upstream commits, not part of this PR. Actual PR diff: 1039 additions / 422 deletions across the wrapper files.

Harsh23Kashyap:
Follow-up on the finalize phase: caught a real bug while reviewing the MMX question.

The earlier comment said the finalize phase was wrapped in `try/except: pass` and called that a "pre-existing pattern, not added by this PR." Went back and looked at it. That pattern was wrong.

`except Exception: pass` swallowed the exception but also left the span open in the SDK. The span was never exported with `.end()` called, so the SDK reported it with an arbitrary end time (or never reported it at all). Real leak.

The generic `_patch_method` in openai.py already had the right pattern: `span.set_status(StatusCode.OK); span.end()` on the failure path. Lifted that into a single helper `_safe_finalize` in `_wrap_utils.py` and applied it across the 6 provider wrappers.

**Fix shape (8 files, +176/-110):**

- `neatlogs/_wrap_utils.py`: new helper `_safe_finalize(span, finalize_fn, *args)`. Calls `finalize_fn`; on exception, sets span status OK and ends it; on second exception, swallows. Never raises.
- `neatlogs/openai.py`, `anthropic.py`, `bedrock.py`, `google_genai.py`, `openrouter.py`, `vertex_ai.py`: all 6 wrappers now use `_safe_finalize` for the post-LLM-call finalize path. The pre-exec `_telemetry_fallback` is unchanged.
- `tests/unit/test_per_call_fail_open.py`: 2 new tests force the finalize callback to throw and assert:
  1. response is returned to the caller
  2. span is in the exporter
  3. span end_time is set
  4. span status is OK

**Verification:**

- 121 unit tests pass (119 + 2 new)
- black + isort + ruff clean on the 8 changed files
- Pre-existing 29 integration failures unchanged (confirmed via `git stash`)
- The `pass` in `_safe()` and the botocore factory guard are untouched (those are not span-finalize paths)

Branch: `fix/per-call-pre-exec-fail-open @ 87037b5`. Pushed.

PR #73, opened 2026-08-18
Finalize active traces during graceful shutdown

tanisha-neatlogs:
## Summary
- track active Neatlogs-owned spans and close them child-first during SDK shutdown
- handle SIGINT/SIGTERM gracefully while preserving existing process signal behavior
- emit the normal root completion marker before provider flush and mark interrupted spans explicitly
- bump the Python SDK to 1.4.21

## Verification
- uv run pytest tests/unit/test_graceful_trace_shutdown.py tests/unit/test_package_version.py -q
- uv run black --check neatlogs/init.py neatlogs/core/span_processor.py tests/unit/test_graceful_trace_shutdown.py

The complete unit suite still requires optional provider extras; collection currently stops at google.genai when those extras are not installed.


neatlogs/init.py
```diff
@@ -127,6 +131,59 @@ def is_debug_enabled() -> bool:
...
+    try:
+        try:
+            reason = signal.Signals(signum).name
+        except ValueError:
+            reason = f"signal-{signum}"
+        shutdown(termination_reason=reason)
+    finally:
+        _signal_shutdown_in_progress = False
+
+    if callable(previous) and previous is not _shutdown_signal_handler:
```
niketan-neatlogs, line 174: Blocking: after calling a previous application handler, this wrapper still always raises `KeyboardInterrupt` or `SystemExit`. That breaks hosts whose SIGTERM handler starts its own graceful shutdown and returns. Callable previous handlers should retain control; apply SDK-owned default/ignored-signal behavior only when the prior disposition is `SIG_DFL`/`SIG_IGN`, or make SDK signal ownership opt-in.

neatlogs/init.py
```diff
@@ -686,10 +749,15 @@ def shutdown(timeout_millis: int = 30000) -> bool:
     except Exception:
         pass
 
+    _restore_shutdown_signal_handlers()
+
     success = True
 
     if _span_processor:
         try:
+            ended = _span_processor.end_active_spans(termination_reason)
```
niketan-neatlogs, line 853: After ending the active spans, shutdown proceeds to the trace provider before the log provider. That lets `neatlogs.trace.complete` reach ingestion/finalization while captured LOG spans are still buffered. Please flush/shutdown logs after ending spans but before flushing/shutting down traces; this should match the existing log-first `flush()` ordering.

neatlogs/core/span_processor.py
```diff
@@ -527,6 +548,60 @@ def on_end(self, span: ReadableSpan) -> None:
...
+
+        clean_reason = str(reason or "shutdown")[:256]
+        ended = 0
+        for span in sorted(active, key=depth, reverse=True):
+            try:
+                if not span.is_recording():
+                    continue
+                span.set_attribute("neatlogs.trace.interrupted", True)
+                span.set_attribute("neatlogs.trace.termination.reason", clean_reason)
+                span.end()
```
niketan-neatlogs, line 684: Please add a `BatchSpanProcessor` regression test around this root end. `span.end()` synchronously invokes this processor, which emits the completion marker before the root reaches the downstream batch processor. With a small batch size, export order becomes child → marker → root, allowing the backend to process completion before any root row is persisted. The root must reach downstream export before its completion marker.

niketan-neatlogs (commented):
GitHub CI is green for lint and Python 3.10–3.13, and the focused graceful-shutdown tests pass. The inline comments cover ordering and host-signal cases that the current suite does not exercise. I would hold the 1.4.21 publication until these are resolved and the post-merge artifact is verified.

neatlogs/init.py
tanisha-neatlogs: I am checking this, didn't fully understand

neatlogs/init.py
tanisha-neatlogs: makes sense. will fix this

neatlogs/core/span_processor.py
tanisha-neatlogs: Cool, will make this change

neatlogs/client.py
```diff
@@ -80,8 +106,15 @@ def __init__(
             except Exception:
                 pass
 
-        self._span_processor = NeatlogsSpanProcessor()
+        self._span_processor = NeatlogsSpanProcessor(
```
niketan-neatlogs, line 109: **Critical — merge blocker.** `own_all_spans=True` makes this Client claim every span created on a caller-supplied tracer provider, including spans owned by the host application. I reproduced a foreign `host-work` span changing from recording to ended after `Client.shutdown()`, with `neatlogs.trace.interrupted=true` and ERROR status. A caller-owned provider may be shared, so shutting down one Client must not terminate unrelated telemetry. Please retain only Neatlogs-scope spans and descendants when the provider is caller-owned, and add a regression proving a foreign root remains recording after Client shutdown.

niketan-neatlogs (commented):
Re-review at `c355a6a`. The earlier callable signal-handler, log-before-trace shutdown, and completion-marker ordering findings are addressed. One new shared-provider lifecycle blocker remains; see the inline comment.

**Non-critical:** none.

Verification: 23 focused shutdown/client tests passed locally, and GitHub lint plus Python 3.10-3.13 tests are green. No API/Zod contract changed.

niketan-neatlogs (approved):

PR #80, opened 2026-08-31
Launch readiness Phase 2: harden shared Python telemetry kernel

tanisha-neatlogs:
## Phase 2 scope

This PR contains the Python shared-SDK-kernel work recorded for launch-readiness Phase 2, plus the safe Python-only portion of Phase 8 that can ship before a backend upload API exists.

- Adds byte-aware OTLP span and log batching, gzip transport, deadline-bounded shutdown/flush behavior, and delivery diagnostics.
- Applies masking to spans and logs through isolated exporters with timeout and failure accounting.
- Preserves multi-choice, streaming, tool-call, typed-media, cancellation, and provider-stream ownership semantics.
- Adds bounded capture with explicit truncation metadata and counters instead of silent/unbounded serialization.
- Removes nested urllib3 read retries; 429 retries honor the exporter deadline and `Retry-After`, while OpenTelemetry remains the sole owner of connection/408/5xx retry behavior.
- Uses workers created during SDK/exporter initialization for bounded Python 3.12+ interpreter-exit cleanup; the `atexit` path does not start a new thread.

## Safe Phase 8 Python SDK scope

- Calculates encoded protobuf size before span and log batching and splits normal traffic conservatively.
- Detects typed image, audio, PDF, binary, and provider media; calculates SHA-256, byte length, MIME type, source, and purpose without fetching remote data or retaining an unbounded decoded copy.
- Removes userinfo, query strings, and fragments from captured remote media locators. When upload is unavailable, large inline media is replaced by typed failure metadata instead of retaining a base64 prefix.
- Bounds span/log names, bodies, status/severity/event fields, resource/record/event/link attributes, mapping keys, strings, and bytes to 100,000 bytes per value and 1 MB of aggregate captured data per telemetry item. Truncation is explicit and collision-safe.
- Adds an internal injectable upload-authority boundary after canonicalization and masking. A bare boolean cannot establish delivery; success requires a structured project-scoped, validated `ready` receipt whose small reference was exported.
- Does not invent an upload endpoint or authentication format. Generic overflow and large typed-media upload remain disabled pending the backend authenticated, project-scoped upload-session contract.

Phase 8 backend upload-session state, validation, quarantine/PII processing, persistence, cleanup, authenticated reads, and end-to-end recovery proof are not implemented in this SDK repository.

## Deployment gate

Gzip remains enabled because the backend is intended to deploy first. Do not release this SDK commit until the default trace and log ingest routes accept and verify gzip-compressed OTLP. The gzip review thread remains open to track that deployment gate.

## Validation

- `uv run pytest -q tests/unit`: 288 passed
- Python 3.12 focused subprocess `atexit` and deadline/concurrent-shutdown regressions: passed
- Python 3.13 built-wheel `atexit` smoke with thread creation forbidden during finalization: passed
- Python 3.12 built-wheel import/client shutdown smoke with `requests==2.31.0` and `urllib3==1.26.20`: passed
- `uv run black --check neatlogs tests/unit`: passed
- `uv run isort --check-only neatlogs tests/unit`: passed
- `uv run python -m compileall -q neatlogs tests/unit`: passed
- `uv build`: wheel and sdist built successfully
- `git diff --check`: passed

### DIFF
--- a/neatlogs/_wrap_utils.py
+++ b/neatlogs/_wrap_utils.py
@@ -21,9 +21,15 @@ from opentelemetry import trace as otel_trace
 from opentelemetry.sdk.trace import TracerProvider
 
 from .constants import DEFAULT_INGEST_ENDPOINT
+from .core.capture import (
+    DEFAULT_MAX_CAPTURE_VALUE_BYTES,
+    UPLOAD_UNAVAILABLE_REASON,
+    bound_text,
+)
 from .core.logger import get_logger
 
 logger = get_logger()
+_MAX_LEGACY_STREAM_CHUNKS = 1024
 
 _wrapper_tracer: Optional[otel_trace.Tracer] = None
 _wrapper_bootstrapped = False
@@ -869,15 +875,48 @@ def get_provider_tracer() -> "_AutoRootTracer":
     return _AutoRootTracer(get_tracer())
 
 
-def serialize(obj: Any, max_length: Optional[int] = None) -> str:
-    """Safe JSON serialization; truncation is only applied when explicitly requested."""
+def serialize(obj: Any, max_length: Optional[int] = DEFAULT_MAX_CAPTURE_VALUE_BYTES) -> str:
+    """Safe JSON serialization with explicit, byte-aware overflow diagnostics."""
     try:
         s = json.dumps(obj, default=str, ensure_ascii=False)
     except (TypeError, ValueError):
         s = str(obj)
-    if max_length is not None and len(s) > max_length:
-        return s[:max_length] + "...[truncated]"
-    return s
+    return s if max_length is None else bound_text(s, max_length)
+
+
+def _accepts_keyword(callback: Callable, name: str) -> bool:
+    try:
+        parameters = inspect.signature(callback).parameters
+    except (TypeError, ValueError):
+        return False
+    parameter = parameters.get(name)
+    return bool(
+        parameter
+        and parameter.kind
+        in {inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.KEYWORD_ONLY}
+    ) or any(item.kind is inspect.Parameter.VAR_KEYWORD for item in parameters.values())
+
+
+class _InterruptedSpanProxy:
+    """Keep legacy finalizers from turning an interrupted stream into success."""
+
+    def __init__(self, span: otel_trace.Span) -> None:
+        self._span = span
+
+    def set_status(self, status_or_status_code, *args: Any, **kwargs: Any) -> None:
+        from opentelemetry.trace import Status, StatusCode
+
+        code = (
+            status_or_status_code.status_code
+            if isinstance(status_or_status_code, Status)
+            else status_or_status_code
+        )
+        if code is StatusCode.OK:
+            return
+        self._span.set_status(status_or_status_code, *args, **kwargs)
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self._span, name)
 
 
 class SyncStreamWrapper:
@@ -894,6 +933,7 @@ class SyncStreamWrapper:
         self._first_chunk_time: Optional[float] = None
         self._incremental = hasattr(finalizer, "on_chunk") and hasattr(finalizer, "finish")
         self._chunks: List[Any] = []
+        self._dropped_chunks = 0
         self._finalized = False
         self._exhausted = False
 
@@ -907,7 +947,9 @@ class SyncStreamWrapper:
                         return
             finally:
                 if not self._finalized:
-                    self.close()
+                    # A loop break abandons this iterator, not the provider-owned
+                    # stream. Finalize telemetry but leave the source resumable.
+                    self._finalize(interrupted=True)
 
         return consume()
 
@@ -924,11 +966,26 @@ class SyncStreamWrapper:
 
         if self._first_chunk_time is None:
             self._first_chunk_time = time.perf_counter()
+        if self._finalized:
+            return chunk
         if self._incremental:
             self._finalizer.on_chunk(self._span, chunk)
-        else:
+        elif len(self._chunks) < _MAX_LEGACY_STREAM_CHUNKS:
             self._chunks.append(chunk)
+        else:
+            self._dropped_chunks += 1
+            self._mark_chunk_overflow()
         return chunk
+
+    def _mark_chunk_overflow(self) -> None:
+        self._span.set_attribute("neatlogs.capture.truncated", True)
+        self._span.set_attribute("neatlogs.capture.truncated_count", self._dropped_chunks)
+        self._span.set_attribute("neatlogs.stream.chunks_dropped", self._dropped_chunks)
+        self._span.set_attribute("neatlogs.capture.overflow.state", "disabled")
+        self._span.set_attribute(
+            "neatlogs.capture.overflow.reason",
+            UPLOAD_UNAVAILABLE_REASON,
+        )
 
     def __enter__(self):
         if hasattr(self._stream, "__enter__"):
@@ -974,7 +1031,21 @@ class SyncStreamWrapper:
         else:
             if interrupted:
                 self._span.set_attribute("neatlogs.stream.cancelled", True)
-            self._finalizer(self._span, self._chunks, elapsed_ms, ttft_ms)
+            if _accepts_keyword(self._finalizer, "interrupted"):
+                self._finalizer(
+                    self._span,
+                    self._chunks,
+                    elapsed_ms,
+                    ttft_ms,
+                    interrupted=interrupted,
+                )
+            else:
+                self._finalizer(
+                    _InterruptedSpanProxy(self._span) if interrupted else self._span,
+                    self._chunks,
+                    elapsed_ms,
+                    ttft_ms,
+                )
 
     def _finalize_error(self, error: BaseException):
         if self._finalized:
@@ -1011,6 +1082,7 @@ class AsyncStreamWrapper:
         self._first_chunk_time: Optional[float] = None
         self._incremental = hasattr(finalizer, "on_chunk") and hasattr(finalizer, "finish")
         self._chunks: List[Any] = []
+        self._dropped_chunks = 0
         self._finalized = False
         self._exhausted = False
 
@@ -1024,7 +1096,9 @@ class AsyncStreamWrapper:
                         return
             finally:
                 if not self._finalized:
-                    await self.aclose()
+                    # Do not transfer ownership of the provider stream to the
+                    # temporary async iterator created for ``async for``.
+                    self._finalize(interrupted=True)
 
         return consume()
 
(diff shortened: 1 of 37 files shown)


neatlogs/init.py
```diff
@@ -447,6 +507,8 @@ def redact(span):
         otlp_exporter = OTLPSpanExporter(
             endpoint=traces_endpoint,
             headers=otlp_headers,
+            compression=Compression.Gzip,
```
niketan-neatlogs, line 512: **Critical: gzip breaks the current default trace receiver.** `ingest.neatlogs.com` resolves to the customer staging lane, and current `origin/staging` (`8956db5`) buffers the wire body and calls `ExportTraceServiceRequest.decode(rawBody)` without handling `Content-Encoding`. This makes every default trace request gzip bytes that the receiver rejects with a non-retryable `400`. The same setting is used by `Client` and wrapper-only bootstrap. Either deploy and verify gzip-aware trace intake before publishing this SDK, or keep trace compression disabled. Please test against the actual trace route middleware rather than a fake server that manually gunzips.

neatlogs/core/deadline.py
```diff
@@ -0,0 +1,34 @@
...
+        return False, TimeoutError("Neatlogs shutdown deadline exceeded")
+    completed: queue.Queue[tuple[bool, Any]] = queue.Queue(maxsize=1)
+
+    def run() -> None:
+        try:
+            completed.put_nowait((True, operation()))
+        except BaseException as exc:
+            completed.put_nowait((False, exc))
+
+    threading.Thread(target=run, name="neatlogs-shutdown", daemon=True).start()
```
niketan-neatlogs, line 85: **High: this cannot run from `atexit` on Python 3.12+.** Both `init()` and `Client` register shutdown functions with `atexit`, and those paths call `bounded_call()`. Python 3.12 and 3.13, which this package supports, raise `RuntimeError` when an exit handler starts a thread. A normal interpreter exit without an earlier explicit shutdown can therefore abort cleanup and lose logs, active spans, or completion markers. Please provide a synchronous exit-handler path, or use a worker created before interpreter finalization, and add subprocess exit tests on 3.12 and 3.13.

neatlogs/core/transport.py
```diff
@@ -0,0 +1,33 @@
...
+    """Retry POST 429/read-timeout failures that OTel Python 1.43 does not.
+
+    The upstream exporter already retries 408, 5xx, and connection failures.
+    Restricting this adapter to 429 and read errors avoids nested retries for
+    the statuses handled by that outer OTLP loop.
+    """
+    retry = Retry(
+        total=2,
+        connect=0,
+        read=2,
```
niketan-neatlogs: **High: these read retries escape the exporter deadline and can duplicate accepted POSTs.** A receiver can read the request and delay its response; urllib3 then retries the POST three times, and the outer OpenTelemetry exporter retries the resulting `ConnectionError` again. A 20 ms exporter timeout reproduced six requests over about 582 ms. Also, the declared `requests>=2.31` range permits urllib3 versions where the adjacent `other` and `allowed_methods` arguments raise `TypeError` during initialization. Please use one absolute-deadline-aware retry layer, avoid ambiguous read retries unless ingest is idempotent, honor bounded `Retry-After`, and declare a compatible urllib3 floor.

neatlogs/_wrap_utils.py
```diff
@@ -856,13 +869,13 @@ def get_provider_tracer() -> "_AutoRootTracer":
     return _AutoRootTracer(get_tracer())
 
 
-def serialize(obj: Any, max_length: int = 100_000) -> str:
-    """Safe JSON serialization with truncation."""
+def serialize(obj: Any, max_length: Optional[int] = None) -> str:
```
niketan-neatlogs: **High: this removes the last per-span payload bound before overflow handling exists.** The base truncated serialized values at 100,000 characters; this now keeps them unbounded. The new byte splitter does not contain that growth because it deliberately forwards an individually oversized span intact, and logs have no byte splitter. Large tool output or inline base64 media can therefore create large in-process copies and then receive a terminal `413`, losing the whole item. Please retain a safe default capture cap until claim-check support ships, or implement single-item overflow handling now and replace inline media bytes with bounded metadata/reference data.

neatlogs/core/masking_exporter.py
```diff
@@ -19,51 +22,149 @@
...
+
+
+def _mask_call(
+    mask: Callable,
+    snapshot: dict[str, Any],
+    timeout_seconds: float,
+    context: MaskContext,
+):
+    result = (
+        mask(snapshot, context) if _accepts_context(mask, snapshot, context) else mask(snapshot)
```
niketan-neatlogs: **Medium: positional arity is not a safe context opt-in.** An existing callback such as `def redact(snapshot, keys=("secret",))` is valid under the old one-argument contract. This check decides it accepts two arguments, replaces `keys` with `MaskContext`, and the callback then fails and drops every affected signal. Please make context support explicit, for example through a dedicated protocol or keyword-only `context` parameter, while preserving one-argument invocation for existing callbacks. Add an optional-second-parameter regression test.

neatlogs/_wrap_utils.py
```diff
@@ -879,55 +892,106 @@ def __init__(self, stream: Any, span: otel_trace.Span, finalizer: Callable):
...
+        def consume():
+            try:
+                while True:
+                    try:
+                        yield self.__next__()
+                    except StopIteration:
+                        return
+            finally:
+                if not self._finalized:
+                    self.close()
```
niketan-neatlogs: **Medium: breaking out of iteration now closes the provider stream.** The temporary generator returned by `__iter__` runs this `finally` on a normal `break`, and `close()` also closes the wrapped source. The base left the source open, so callers could resume it later; the head makes the next read fail. Instrumentation should not change provider-stream ownership. Please finalize telemetry without closing the source on loop abandonment, and reserve source closure for explicit `close()`, context-manager exit, or normal exhaustion. Add a break-then-resume test.

neatlogs/core/choice_accumulator.py
```diff
@@ -0,0 +1,311 @@
...
+        for choice_index in indexes:
+            choice = self.choices[choice_index]
+            prefix = f"neatlogs.llm.output_messages.{choice_index}"
+            span.set_attribute(f"{prefix}.role", choice.role)
+            content = "".join(choice.content)
+            if content:
+                span.set_attribute(f"{prefix}.content", content)
+            reasoning = "".join(choice.reasoning)
+            if reasoning:
+                span.set_attribute(f"{prefix}.reasoning", reasoning)
```
niketan-neatlogs: **High: this field bypasses the backend PII contract.** The backend selects indexed model reasoning only as `.thinking`, and the existing provider adapters plus the Go SDK use that spelling. Emitting `.reasoning` means `reasoning_content` is neither server-side redacted nor recognized by consumers that render model thinking. Please emit `${prefix}.thinking` and add streamed and non-streamed contract tests against the backend PII selector.

neatlogs/_wrap_utils.py
```diff
@@ -879,55 +892,106 @@ def __init__(self, stream: Any, span: otel_trace.Span, finalizer: Callable):
...
+        if self._incremental:
+            self._finalizer.finish(
+                self._span,
+                elapsed_ms,
+                ttft_ms,
+                interrupted=interrupted,
+            )
+        else:
+            if interrupted:
+                self._span.set_attribute("neatlogs.stream.cancelled", True)
```
niketan-neatlogs, line 1095: **Medium: cancellation status is overwritten for every legacy stream finalizer.** On an early close this branch sets `neatlogs.stream.cancelled=true`, then callbacks such as OpenAI Responses, Anthropic, and Google GenAI unconditionally mark the span `OK`. A cancelled stream therefore looks successful. Please pass an explicit terminal reason or `interrupted` flag into legacy finalizers so cancellation remains `UNSET`, and cover early close for at least one callback-based provider.

neatlogs/core/byte_limited_exporter.py
```diff
@@ -0,0 +1,71 @@
...
+            current.append(span)
+            current_bytes += span_bytes
+
+        if current:
+            batches.append(current)
+
+        for batch in batches:
+            if self._inner.export(batch) is not SpanExportResult.SUCCESS:
+                if self._diagnostics is not None:
+                    self._diagnostics.record_export_failure("span", len(batch))
```
niketan-neatlogs: **Medium: the loss counter omits every unattempted tail batch.** Export returns immediately after one split batch fails, so all later batches are also undelivered, but this records only the current batch. With three one-span batches and failure on the second export, two spans are lost while the diagnostic reports one. Please count the failed batch plus all remaining unattempted batches, excluding only batches already exported successfully, and add first/middle/last failure tests.

niketan-neatlogs (changes requested):
Requesting changes after a full review of the 24-file diff and its surrounding SDK/backend contracts. The inline findings include a release-blocking incompatibility with the current default ingest route, Python 3.12+ shutdown failure, masking and PII contract regressions, and delivery/streaming correctness issues. Focused changed-area tests pass (44 tests), so these are coverage gaps rather than existing test failures. No application API/Zod contract, database schema, migration, or backfill changes are present.

neatlogs/_wrap_utils.py
tanisha-neatlogs: Fixed in 5bc4868. The wrappers pass interrupted explicitly to in-tree legacy finalizers, all callback-based provider finalizers preserve UNSET when interrupted, and a compatibility proxy prevents old callback signatures from overwriting cancellation with OK. Added an OpenAI Responses early-close regression.

neatlogs/core/masking_exporter.py
tanisha-neatlogs: Fixed in 5bc4868. MaskContext is passed only through an explicit keyword-only context parameter. Existing callbacks with unrelated optional positional parameters retain their original defaults. Added regressions for both the legacy optional-positional form and keyword-only opt-in.

neatlogs/core/byte_limited_exporter.py
tanisha-neatlogs: Fixed in 5bc4868. A failed split export now counts the failed batch plus every remaining unattempted batch. Added first/middle/last failure regressions for spans and logs.

neatlogs/_wrap_utils.py
tanisha-neatlogs: Fixed in 5bc4868. Abandoning the temporary iterator now finalizes telemetry as interrupted without closing the provider-owned stream; only explicit close/context-manager exit closes it. Added a break/iterator-close then resume-source regression.

neatlogs/_wrap_utils.py
tanisha-neatlogs: Fixed in 5bc4868. Default serialization is bounded again at 100,000 UTF-8 bytes with an explicit original-size/digest/backend-dependency marker; final post-mask span and log capture also has a 1 MB aggregate string budget and diagnostics. Legacy stream buffering is bounded. Individually oversized spans/logs now cross a disabled/injectable post-mask authority seam and fail with counters instead of being forwarded into a terminal 413. Large typed media reports the unavailable backend upload contract explicitly.

neatlogs/core/choice_accumulator.py
tanisha-neatlogs: Fixed in 5bc4868. Indexed reasoning content now emits as neatlogs.llm.output_messages.N.thinking. Added non-streamed and streamed contract assertions that .reasoning is absent.

neatlogs/core/transport.py
tanisha-neatlogs: Fixed in 5bc4868. urllib3 adapter retries are disabled, so ambiguous read timeouts cannot replay accepted POSTs or nest beneath the OTel retry loop. The dedicated session retries only definite 429 responses, uses the remaining exporter timeout as an absolute budget, honors bounded Retry-After, and avoids version-sensitive urllib3 Retry arguments. Covered read-timeout, deadline, 429, 503, shutdown, and requests 2.31/urllib3 1.26 compatibility.

neatlogs/core/deadline.py
tanisha-neatlogs: Fixed in 5bc4868. Both process-default and Client atexit handlers now select a synchronous shutdown path; bounded worker threads remain available only for explicit runtime shutdown. Added subprocess regressions that replace Thread.start with a failure and verify cleanup completes, exercised locally on Python 3.12 and 3.13.

neatlogs/init.py
tanisha-neatlogs: Gzip remains enabled by explicit deploy order: the backend must deploy and verify gzip-aware default trace and log intake before this SDK commit is released. I documented that release gate in README.MD and the PR body. Leaving this thread unresolved until the backend route is deployed and verified, as requested.

tanisha-neatlogs: Accepted as a backend-first release gate. The backend development line now implements bounded gzip/deflate decoding in the actual `/v1/traces` middleware, with route-level tests for successful gzip intake and decompressed-size rejection. Backend deployment will occur before this Python SDK version is released, so compression remains enabled and the dependency is recorded in the expanded PR description.

niketan-neatlogs (approved):

PR #81, opened 2026-09-01
feat: add Python Doctor v2 diagnostics

shyam-neatlogs:
## What and why

Adds Python Doctor v2 diagnostics at the final normalized/masked exporter boundary. Probe mode uses the existing trace pipeline, not a diagnostic receipt service:

`POST /v1/traces` → Kafka/finalizer → exact authenticated `GET /api/traces/v3/:traceId`.

## Shipped

- bounded Doctor-only post-mask capture and stable reason/remediation codes
- `neatlogs doctor` and `neatlogs-doctor` CLI entry points
- exact four-span controlled fixture and deterministic semantic digest
- exact finalized trace read-back with hierarchy, I/O, metadata, and numeric-token checks
- shared rich fixture for choices, tool requests, streaming, and payload references
- real public Google GenAI-compatible wrapper → normalizer → masked capture coverage
- ordinary/Doctor initialization isolation in both orders
- canonical `neatlogs.span.kind` and exact/wildcard Doctor contract registration
- fail-closed credential, endpoint, auth, timeout, and mismatch handling

## Validation

- [x] full suite: 410 tests passed
- [x] Black: 139 files clean
- [x] isort passed
- [x] wheel and sdist build passed
- [x] shared schema hash: `sha256:50bbd9f1e6eaa6c83f08dcb84da3a98867c962fc8c4e1edd629da561fe5fe5a8`
- [x] shared simple digest: `sha256:7163d2de42c4165f3ae552279fdde2ec0839413ce608c6e5d71f3fb532df319b`
- [x] final-head local backend probe passed: `917fb1fb2e8aa4e1a1faa748e47fca55`
- [x] missing/invalid credentials and invalid endpoint return exit 3 with stable reasons
- [ ] final CI completion
- [ ] reviewer re-review
- [ ] publish package after merge

## Intentionally excluded

Generated artifacts, credentials, obsolete Doctor session APIs, and unrelated provider changes.


shyam-neatlogs:
Fixed the CI-only fixture failure: the cross-SDK golden envelope is now vendored under `tests/fixtures/doctor-v2` instead of referencing a developer-local Chotu path. Local CI-equivalent validation remains 301 passed; lint/format checks pass.

neatlogs/doctor_v2.py
```diff
@@ -0,0 +1,753 @@
...
+    "QUEUE_SATURATED": "INCREASE_OR_DRAIN_QUEUE",
+    "EXPORT_RETRY_EXHAUSTED": "CHECK_TRANSPORT",
+    "FLUSH_TIMEOUT": "INCREASE_FLUSH_BUDGET",
+    "MASKING_FAILED_CLOSED": "FIX_MASK_CALLBACK",
+    "LOCAL_ENVELOPE_VALID": "NONE",
+    "STAGE_PENDING": "WAIT_FOR_RECEIPT",
+    "DIAGNOSTIC_EXPIRED": "CREATE_NEW_SESSION",
+    "DIAGNOSTIC_NOT_VISIBLE": "CONTACT_SUPPORT",
+    "DIGEST_MISMATCH": "CONTACT_SUPPORT",
+}
```
tanisha-neatlogs, line 91: How exhaustive is this list?

tanisha-neatlogs (changes requested):
Requesting changes because the Python Doctor v2 implementation does not exercise the SDK or hosted ingestion path and adds unbounded-by-content diagnostic retention to every normal export.

## Blockers

1. **The probe bypasses the Python SDK entirely.** `neatlogs/doctor_v2.py` constructs a synthetic dictionary and posts only digest/version/trace ID to the diagnostic session. No SDK span processor, exporter, OTLP ingestion, project routing, persistence, or finalization is exercised. The probe therefore cannot establish end-to-end health. Generate the trace through an isolated real SDK pipeline, export it normally with exact correlation, flush, and poll persisted/finalized readback.

2. **Doctor capture runs on every normal export.** `neatlogs/core/masking_exporter.py` invokes the global capture path for production spans. The store limits trace count but not spans or total bytes, duplicates parsed JSON and original attributes, and survives normal initialization/shutdown unless explicitly cleared by Doctor/tests. This creates memory/privacy retention unrelated to Doctor. Install capture only for Doctor, enforce byte/span bounds, and clear it deterministically.

3. **Doctor maintains a second attribute schema outside `neatlogs/config/attribute-mapping.json`.** Hand-written readers include unregistered names such as `gen_ai.response.choices`, `neatlogs.llm.choices`, `gen_ai.assistant.tool_calls`, aggregate `neatlogs.llm.tool_calls`, and both stream-fragment names. The registry instead defines `gen_ai.client.generation.choices` -> `neatlogs.llm.generation_choices`, indexed tool calls, and the existing streaming representation. Use the registry as the source of truth. Newer valid OpenTelemetry aliases must be added centrally rather than only inside Doctor.

4. **The cross-language projection is inconsistent.** Python disables completion markers and produces two local spans, while TypeScript/Go include the marker and produce three. Full attributes and volatile timing also conflict with the shared exclusion policy. Define and test a single language-neutral projection over actual captured SDK output.

5. **The required hosted diagnostics endpoint is not present in merged backend code.** Shipping this CLI before the server contract exists makes probe mode unavailable by construction.

## Required regression coverage

- ordinary SDK exports do not enter a Doctor retention store;
- byte/span caps and cleanup across shutdown/reinitialization;
- real supported wrapper output for streaming, choices, tools, and payload handling;
- actual SDK export -> authenticated backend -> exact finalized readback;
- identical canonical projection and digest across all three SDKs.

The branch is based on current `main`; I found no direct revert of prior lifecycle/retriever changes or new threading defect. The full unit suite (301 tests), formatting checks, and package build passed, but the current tests use synthetic envelopes/mocked receipts and do not cover these blockers.

neatlogs/doctor_v2.py
shyam-neatlogs: The mapping is exhaustive for every reason code currently emitted by the local validator and backend probe. Unknown future or server-provided reason codes now fail safely to CONTACT_SUPPORT instead of returning an empty or misleading remediation; the regression is covered by test_unknown_future_reason_uses_safe_support_remediation. The final full suite passes 410 tests.

tests/unit/test_doctor_v2.py
```diff
@@ -0,0 +1,1193 @@
...
+            "protocolVersion": "v1",
+            "state": "succeeded",
+            "currentStage": "finalized",
+            "lastSuccessfulStage": "finalized",
+            "retryable": False,
+            "stages": [{"lastObservedAt": "must-not-survive"}],
+        }
+        return _json_response(payload)
+
+    monkeypatch.setattr("neatlogs.doctor_v2.requests.get", get)
```
tanisha-neatlogs, line 574: [Blocker] The earlier review required an actual SDK export → authenticated backend → exact finalized readback regression. This test replaces the read API and injects an in-memory exporter, so CI can still pass when project routing, PII/Kafka processing, finalization, or hosted readback is broken. Keep the unit test, but add a protected integration/canary job that installs the packed wheel, exports to a deployed backend, and reads back the exact four-span trace before treating that review item as closed.

shyam-neatlogs: Implemented the packed-wheel canary in `.github/workflows/doctor-e2e.yml`. The latest run passed against the deployed backend: https://github.com/neatlogs/neatlogs/actions/runs/33836606836. It installed the built wheel, exported through `/v1/traces`, and asserted exact trace-ID readback, finalized=true, four spans, one root, zero duplicates, hierarchy/attributes/input-output/metadata, and numeric tokens. Sanitized JSON is attached as the `python-doctor-e2e-*` artifact. Credentials are stored as a GitHub repository secret

tanisha-neatlogs (approved):

PR #83, opened 2026-09-02
Phase 6: coalesce exact SDK prompt fetches

niketan-neatlogs:
## Problem

The Python SDK resolved latest and exact managed-prompt versions by listing only the first 100 history rows. Concurrent cold-cache callers also performed duplicate HTTP requests for the same prompt key.

## What changed

- Uses the exact backend prompt endpoint for label, version, and latest selectors.
- Uses the same exact lookup before update-label, delete, and remove-tag mutations instead of listing prompt history.
- Coalesces concurrent synchronous cold misses per cache key.
- Coalesces asynchronous cold misses into one shielded task so cancelling one caller does not cancel the shared fetch.
- Propagates one shared fetch failure to every waiting synchronous caller and clears the flight for a later retry.

## Resulting behavior

Prompt-history size no longer affects correctness. A burst of callers asking for the same uncached prompt produces one backend request instead of one request per caller.

## Contract and rollout notes

- No public Python method signature changed.
- No migration, data rewrite, deployment, or live database query.
- Requires the companion backend exact-selector route to be released first.

## Test plan

- Complete SDK unit suite with the OpenAI extra: 388 passed, 1 skipped.
- Focused exact-fetch and coalescing suite: 9 passed.
- `black --check`
- `isort --check-only`
- `python -m compileall -q neatlogs`
- `git diff --check`

## Tracking

- [Linear NL-1728](https://linear.app/neatlogs/issue/NL-1728/phase-6-frontend-and-ai-service-query-review)
- [Living Phase 6 status](https://app.notion.com/p/3c230dad43af8118a78fd29896c44ffd)

## Mandatory self-review update — 2 September 2026

Both required review workflows were applied to the exact SDK diff. The review found that switching latest/version resolution to the exact endpoint changed missing-prompt failures from the exported `PromptNotFoundError` to generic `PromptApiError`, and that keeping the new concurrency state inside the already-large client would leave it at the 1,000-line boundary. Commit `e0d4f0f` preserves `PromptNotFoundError` for exact 404s and moves synchronous/asynchronous single-flight ownership into a focused private module; `client.py` is 987 lines.

Final verification: 13 focused exact-fetch/single-flight tests passed; the full SDK suite passed 392 tests with 1 skip; Black, isort, compileall, and diff checks passed. No API signature, migration, deployment, or live backend/database operation was performed.

### DIFF
--- a/neatlogs/prompt/_singleflight.py
+++ b/neatlogs/prompt/_singleflight.py
@@ -1,0 +1,77 @@
+from __future__ import annotations
+
+import asyncio
+import threading
+from dataclasses import dataclass, field
+from typing import Awaitable, Callable, Dict, Generic, Optional, TypeVar, cast
+
+T = TypeVar("T")
+
+
+@dataclass
+class _SyncFlight(Generic[T]):
+    event: threading.Event = field(default_factory=threading.Event)
+    error: Optional[BaseException] = None
+    result: Optional[T] = None
+    has_result: bool = False
+
+
+class SyncSingleFlight(Generic[T]):
+    """Run one synchronous operation per key and share its outcome with waiters."""
+
+    def __init__(self) -> None:
+        self._lock = threading.Lock()
+        self._flights: Dict[str, _SyncFlight[T]] = {}
+
+    def run(self, key: str, operation: Callable[[], T]) -> T:
+        with self._lock:
+            flight = self._flights.get(key)
+            leader = flight is None
+            if flight is None:
+                flight = _SyncFlight()
+                self._flights[key] = flight
+
+        if not leader:
+            flight.event.wait()
+            if flight.error is not None:
+                raise flight.error
+            if not flight.has_result:
+                raise RuntimeError("Single-flight operation completed without a result")
+            return cast(T, flight.result)
+
+        try:
+            flight.result = operation()
+            flight.has_result = True
+            return cast(T, flight.result)
+        except BaseException as error:
+            flight.error = error
+            raise
+        finally:
+            flight.event.set()
+            with self._lock:
+                if self._flights.get(key) is flight:
+                    self._flights.pop(key, None)
+
+
+class AsyncSingleFlight(Generic[T]):
+    """Run one event-loop task per key without transferring caller cancellation."""
+
+    def __init__(self) -> None:
+        self._tasks: Dict[str, asyncio.Task[T]] = {}
+
+    async def run(self, key: str, operation: Callable[[], Awaitable[T]]) -> T:
+        task = self._tasks.get(key)
+        if task is None:
+            task = asyncio.create_task(operation())
+            self._tasks[key] = task
+
+            def clear_flight(completed: asyncio.Task[T]) -> None:
+                if self._tasks.get(key) is completed:
+                    self._tasks.pop(key, None)
+                # A shielded task can outlive every cancelled waiter. Observe its
+                # terminal exception so asyncio does not report it as unhandled.
+                if not completed.cancelled():
+                    completed.exception()
+
+            task.add_done_callback(clear_flight)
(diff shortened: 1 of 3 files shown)


neatlogs/prompt/_singleflight.py
```diff
@@ -0,0 +1,77 @@
...
+    def run(self, key: str, operation: Callable[[], T]) -> T:
+        with self._lock:
+            flight = self._flights.get(key)
+            leader = flight is None
+            if flight is None:
+                flight = _SyncFlight()
+                self._flights[key] = flight
+
+        if not leader:
+            flight.event.wait()
```
tanisha-neatlogs, line 35: Please remove this unbounded waiter when rebasing onto merged #84. #84 already coalesces synchronous cold misses and bounds follower waits using the client transport deadline; event.wait() here can block forever if the leader hangs.

neatlogs/prompt/_singleflight.py
```diff
@@ -0,0 +1,77 @@
...
+class AsyncSingleFlight(Generic[T]):
+    """Run one event-loop task per key without transferring caller cancellation."""
+
+    def __init__(self) -> None:
+        self._tasks: Dict[str, asyncio.Task[T]] = {}
+
+    async def run(self, key: str, operation: Callable[[], Awaitable[T]]) -> T:
+        task = self._tasks.get(key)
+        if task is None:
+            task = asyncio.create_task(operation())
```
tanisha-neatlogs, line 65: Please use the client-owned task lifecycle from merged #84 rather than this standalone map. #84 shields a shared request from one caller cancellation while still tracking, cancelling, and awaiting it during AsyncPromptClient.close(); this task can otherwise outlive shutdown.

neatlogs/prompt/client.py
```diff
@@ -259,23 +293,22 @@ def _fetch_prompt(
...
-        if version is not None:
-            for item in items:
-                if item.get("version") == version:
-                    return PromptHandle(_normalize_prompt_object(item))
-            raise PromptNotFoundError(f"Prompt '{name}' version {version} not found")
-
-        latest = max(items, key=lambda x: x.get("createdAt") or x.get("created_at") or "")
-        return PromptHandle(_normalize_prompt_object(latest))
+            params = {"label": label}
+        elif version is not None:
```
tanisha-neatlogs, line 300: Keep this exact-selector optimization only after neatlogs/neatlogs-app#1239 is reconciled, merged, and deployed. The backend after merged neatlogs/neatlogs-app#1259 still supports label on this fetch route, not version or latest=true. Until then, retain #84 pagination, which is correct for histories beyond 100 versions.

neatlogs/prompt/client.py
```diff
@@ -539,7 +548,10 @@ def _request_json(
 
         if response.status_code >= 400:
             body = _safe_response_text(response)
```
tanisha-neatlogs, line 550: Please remove response-body text from this public error when rebasing onto merged #84, and make the same change in the async path below. Backend or proxy bodies can contain prompt data; #84 intentionally exposes only safe method/path/status/request-ID metadata.

niketan-neatlogs:
Closing this PR because merged SDK PR #84 supersedes its safe, deployable scope with stronger lifecycle and privacy guarantees: bounded synchronous coalescing, client-owned asynchronous task cancellation/cleanup, paginated prompt-history lookups, and redacted transport errors.

The superseding implementation was verified on current main with the full feasible SDK suite: 410 passed and 1 skipped, plus 31 focused prompt-cache tests.

Exact version/latest selector requests remain intentionally deferred until neatlogs/neatlogs-app#1239 is reconciled, merged, deployed, and its backend contract is revalidated. The currently deployed-compatible fetch contract supports exact label lookup, so merging this branch now would introduce unsupported requests.

PR #87, opened 2026-09-05
Enable explicit HTTP client instrumentation

harshk-neatlogs:
## Summary

- Enable HTTP client instrumentation when `requests`, `httpx`, `urllib3`, or `aiohttp` is explicitly requested.
- Bind each instrumentor to the Neatlogs tracer provider.
- Keep Neatlogs telemetry requests filtered and continue dropping unrelated rootless HTTP noise.
- Cover nested requests, failures, shutdown, reinitialization, and duplicate prevention.

## Why

The public instrumentation keys were accepted, but the HTTP setup path returned without installing the requested instrumentors.

## Testing

- `uv run pytest -q`: 454 passed, 1 pre-existing skip

## Merge note

This PR is independent, but it touches shared instrumentation setup and may need a mechanical rebase if another provider-routing PR merges first.

### DIFF
--- a/tests/unit/test_http_instrumentation_e2e.py
+++ b/tests/unit/test_http_instrumentation_e2e.py
@@ -1,0 +1,98 @@
+import asyncio
+import threading
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+
+import aiohttp
+import httpx
+import pytest
+import requests
+import urllib3
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
+
+from neatlogs.instrumentation.manager import InstrumentationManager
+
+
+class _Handler(BaseHTTPRequestHandler):
+    def do_GET(self):
+        body = b"ok"
+        self.send_response(200)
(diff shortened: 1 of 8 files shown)


tests/unit/test_http_instrumentation_e2e.py
```diff
@@ -0,0 +1,98 @@
...
+        with tracer.start_as_current_span("workflow") as root:
+            _request(library, local_http_url)
+
+        children = [
+            span
+            for span in exporter.get_finished_spans()
+            if span.parent and span.parent.span_id == root.context.span_id
+        ]
+        assert len(children) == 1
+        assert children[0].kind.name == "CLIENT"
```
tanisha-neatlogs, line 93: **Blocker:** This test bypasses `NeatlogsSpanProcessor` and verifies only the OTel `CLIENT` kind. Running the request through the normal Neatlogs pipeline produces an HTTP span without canonical `input.value` or `output.value`. The Wizard requires I/O on every persisted non-LOG span, so enabling this integration makes an otherwise valid setup fail verification. Please test through the public SDK pipeline and assert safe canonical I/O—for example method plus sanitized URL as input and status as output, without capturing sensitive bodies.

tanisha-neatlogs (changes requested):
The explicit HTTP activation fix works, but the emitted span contract is not complete yet.

tanisha-neatlogs (approved):

PR #88, opened 2026-09-05
Use native Pydantic AI instrumentation

harshk-neatlogs:
## Summary

- Configure Pydantic AI through its native `Agent.instrument_all(...)` API.
- Install the OpenInference processor on the private Neatlogs tracer provider.
- Preserve and restore prior framework instrumentation during shutdown.
- Cover semantic spans, private/global provider isolation, lifecycle, and repeated initialization.

## Why

The previous path expected a `PydanticAIInstrumentor` class that the supported Pydantic AI integration does not provide, leaving the advertised integration inactive.

## Testing

- `uv run pytest -q`: 457 passed, 1 pre-existing skip

## Merge note

This PR is independent, but it touches shared provider-isolation setup and may need a mechanical rebase if another provider-routing PR merges first.


neatlogs/instrumentation/manager.py
```diff
@@ -38,6 +68,53 @@ def __init__(
...
+            self._pydantic_ai_processor = processor
+
+            settings = InstrumentationSettings(
+                tracer_provider=provider_for_native_instrumentation(
+                    self.provider,
+                    module_prefixes=("pydantic_ai",),
+                ),
+                include_content=True,
+            )
+            Agent.instrument_all(settings)
```
tanisha-neatlogs, line 125: **Blocker:** `Agent.instrument_all(settings)` installs a process-global default bound permanently to this manager's provider, while `set_native_auto_instrumentation(True)` makes every `neatlogs.wrap(agent)` skip its Client-aware wrapper. I reproduced default `init()` plus `Client.activate()` sending the Pydantic AGENT/LLM/TOOL children to the default exporter while the WORKFLOW root went to the Client exporter, splitting one trace across projects. Please route native spans through the active Client provider—or retain Client-local wrapper ownership—and add a two-exporter regression test.

pyproject.toml
```diff
@@ -155,7 +155,8 @@ portkey = [
     "portkey-ai>=1.0.0",
 ]
 pydantic-ai = [
-    "pydantic-ai>=0.0.9",
+    "pydantic-ai>=0.7.5",
+    "openinference-instrumentation-pydantic-ai[instruments]>=0.1.9",
```
tanisha-neatlogs: **Compatibility blocker:** this range permits `openinference-instrumentation-pydantic-ai==0.1.16`, but the new hierarchy test fails with that supported version because the expected CHAIN span is absent; the locked 0.1.20 environment passes. Please either raise the minimum to the first version that provides the required hierarchy or support the older output, and run the integration test at the declared minimum.

tanisha-neatlogs (changes requested):
Native Pydantic AI instrumentation fixes the simple global-init case, but provider isolation and the declared dependency range still fail supported configurations.

tanisha-neatlogs (approved):

PR #89, opened 2026-09-05
Keep Google ADK spans on the SDK provider

harshk-neatlogs:
## Summary

- Route Google ADK instrumentation through the private Neatlogs OpenInference provider.
- Use the same provider path for automatic activation and explicit runner wrapping.
- Remove shared request-message state that could cross concurrent runs.
- Prevent duplicate semantic spans and restore installed instrumentation during shutdown.

## Why

ADK runs could produce the workflow wrapper while losing the expected agent, model, and tool child spans because framework instrumentation was not consistently bound to the SDK provider.

## Testing

- `uv run pytest -q`: 454 passed, 1 pre-existing skip
- Google ADK hierarchy, tools, errors, concurrency, masking, lifecycle, and duplicate-prevention tests passed

## Merge note

This PR is independent, but it touches shared provider-isolation setup and may need a mechanical rebase if another provider-routing PR merges first.

### DIFF
--- a/neatlogs/google_adk.py
+++ b/neatlogs/google_adk.py
@@ -11,29 +11,21 @@ Usage:
 import time
 from typing import Any
 
-from opentelemetry import context as otel_context
 from opentelemetry.trace import StatusCode
 
 from ._wrap_utils import attach_as_current, detach, get_tracer, serialize
-
-_ADK_HOOKS_INSTALLED = False
-
-# The inner `generate_content` span (trace_inference_result) only receives the
-# response, not the request. We stash the most recent request's structured messages
-# in a plain module global so the inner span can reuse them. Calls are effectively
-# sequential per agent turn, and we set it both at use_inference_span entry (which
-# DOES receive llm_request) and in trace_call_llm — whichever fires first wins.
-_LAST_REQUEST_MESSAGES: list = []
 
 
 def wrap_google_adk(runner: Any) -> Any:
     """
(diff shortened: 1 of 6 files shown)


neatlogs/google_adk.py
```diff
@@ -11,29 +11,21 @@
...
-    user-facing WORKFLOW span, and installs hooks on ADK's own telemetry helpers
-    so its native per-call spans carry neatlogs I/O + kind.
+    user-facing WORKFLOW span and activates ADK's official OpenInference
+    instrumentor on the current Neatlogs tracer provider.
     Returns the same runner instance.
     """
-    _install_adk_telemetry_hooks()
+    from .init import _instrument_library
+
+    _instrument_library("google_adk")
```
tanisha-neatlogs: **Blocker:** this resolves only the current global instrumentation manager at wrapper-construction time. If the runner is wrapped before `neatlogs.init()`, it returns false and semantic instrumentation is never retried. With default `init()` plus `Client.activate()`, ADK's LLM/AGENT/CHAIN children remain on the default provider while the wrapper root goes to the Client provider, splitting the trace across projects. Please defer provider resolution to invocation/initialization and add both wrap-before-init and two-exporter Client regressions.

tanisha-neatlogs (changes requested):
The official ADK instrumentor restores semantic children in the ordinary global-init path, but the activation point breaks supported lifecycle and Client routing cases.

tanisha-neatlogs (approved):

PR #90, opened 2026-09-05
Route Strands telemetry through the SDK provider

harshk-neatlogs:
## Summary

- Install the Strands OpenInference processor on the private Neatlogs tracer provider.
- Route the Strands tracer singleton and explicitly wrapped agents through that provider.
- Restore prior framework state on shutdown and avoid duplicate processors, hooks, and tracers.
- Cover pre-created agents, semantic hierarchy, concurrency, errors, lifecycle, and repeated wrapping.

## Why

Strands runs could retain the outer workflow while losing agent, chain, model, and tool child spans because native framework telemetry was not connected to the SDK provider.

## Testing

- `uv run pytest -q`: 452 passed, 1 pre-existing skip
- The five Python fixes also passed together: 488 passed, 1 pre-existing skip

## Merge note

This PR is independent, but it touches shared instrumentation and dependency files and may need a mechanical rebase if another provider-routing PR merges first.

### DIFF
--- a/neatlogs/init.py
+++ b/neatlogs/init.py
@@ -5,6 +5,7 @@ Neatlogs SDK.
 import atexit
 import functools
 import hashlib
+import importlib.util
 import json
 import math
 import os
@@ -90,6 +91,15 @@ _session_config = {
 def is_debug_enabled() -> bool:
     """Return True if neatlogs was initialized with debug=True."""
     return _debug_mode
+
+
+def _instrument_library(library: str) -> bool:
+    """Activate one library through the current default instrumentation manager."""
+    manager = _instrumentation_manager
+    if manager is None:
+        return False
+    manager.instrument(libraries=[library])
+    return library in manager.instrumented
 
 
 def _trace_sampler(sample_rate: float) -> ParentBased:
@@ -540,6 +550,19 @@ def init(
     from ._wrap_utils import set_neatlogs_provider
 
     set_neatlogs_provider(provider)
+
+    # Strands converts its native GenAI spans in an on_end processor. It must run
+    # before Neatlogs' normalizer and exporter see those spans.
+    if importlib.util.find_spec("strands") is not None:
+        try:
+            from .strands import instrument_strands, prepare_strands
+
+            prepare_strands(provider)
+            if instrumentations and "strands" in instrumentations:
+                instrument_strands(provider)
+        except Exception as exc:
+            if debug:
+                logger.debug("Could not prepare Strands instrumentation: %s", exc)
 
     # NeatlogsSpanProcessor: pure pre-processing (attribute normalization + file logging)
     global _span_processor
--- a/neatlogs/init.py
+++ b/neatlogs/init.py
@@ -5,7 +5,6 @@ Neatlogs SDK.
 import atexit
 import functools
 import hashlib
-import importlib.util
 import json
 import math
 import os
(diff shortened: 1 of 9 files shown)


neatlogs/strands.py
```diff
@@ -1,188 +1,138 @@
...
-single class-level hook on Strands' own ``Tracer._add_event`` chokepoint: whenever
-Strands records a gen_ai message/choice event, we ALSO set input.value/output.value
-(+ span kind) on the same span. We do NOT create our own spans — Strands' native
-tracing stays the source of truth; we only enrich it.
-"""
+    prepare_strands(provider)
+    oi_provider = provider_for_openinference(provider)
 
-from typing import Any
+    global _ACTIVE_PROVIDER, _ACTIVE_TRACER, _PREVIOUS_TRACER
```
tanisha-neatlogs: **Blocker:** `_ACTIVE_PROVIDER`/`_ACTIVE_TRACER` are process-global. I reproduced default `init()` plus `Client.activate()` sending Strands semantic spans through the default exporter while the Client WORKFLOW used a different trace/provider. This loses the hierarchy in both projects. Provider selection must follow the active Client/run context; please add a two-exporter regression.

neatlogs/strands.py
```diff
@@ -1,188 +1,138 @@
...
-    """
-    _install_event_hook()
-    try:
-        setattr(agent, "_neatlogs_patched", True)
-    except Exception:
-        pass
+    """Route an existing Strands agent through Neatlogs and return it unchanged."""
+    from .init import _instrument_library
+
+    _instrument_library("strands")
```
tanisha-neatlogs: **Blocker:** wrapper activation is attempted only here. If `strands_hooks(agent)` runs before `neatlogs.init()`, `_instrument_library()` returns false and this agent keeps its old tracer; initialization never revisits it. Please defer binding until initialization/invocation and add a wrap-before-init regression.

neatlogs/strands.py
```diff
@@ -1,188 +1,138 @@
...
-    _HOOK_INSTALLED = True
+def _remember_agent(agent: Any, previous_tracer: Any) -> None:
+    for reference, _ in _WRAPPED_AGENTS:
+        if reference() is agent:
+            return
+    try:
+        reference = weakref.ref(agent)
+    except TypeError:
+        reference = lambda: agent
+    _WRAPPED_AGENTS.append((reference, previous_tracer))
```
tanisha-neatlogs: **Memory regression:** dead weak references are appended to `_WRAPPED_AGENTS` and never pruned until global shutdown, while each tuple also retains `previous_tracer`. A service that creates and wraps agents dynamically grows this list for its entire lifetime. Please use a weak-key structure or prune dead entries during registration.

neatlogs/init.py
```diff
@@ -541,6 +551,19 @@ def redact(span):
 
     set_neatlogs_provider(provider)
 
+    # Strands converts its native GenAI spans in an on_end processor. It must run
+    # before Neatlogs' normalizer and exporter see those spans.
+    if importlib.util.find_spec("strands") is not None:
+        try:
+            from .strands import instrument_strands, prepare_strands
+
+            prepare_strands(provider)
```
tanisha-neatlogs: **Opt-in regression:** merely having `strands` installed now adds its conversion processor to every Neatlogs provider, even when `instrumentations` does not include Strands. That adds global processing/third-party mutation to unrelated spans and contradicts explicit instrumentation selection. Please install the processor only when Strands is selected, while preserving the required processor ordering.

tanisha-neatlogs (changes requested):
The native Strands conversion works in the locked single-provider test, but it introduces provider-routing, lifecycle, and retention regressions.

tanisha-neatlogs (approved):

PR #92, opened 2026-09-07
fix(doctor): report actionable probe failures

shyam-neatlogs:
## What changed

- Classify authentication, timeout, pipeline, HTTP, connection, and invalid-readback failures separately.
- Preserve only allowlisted ingestion receipt fields in Doctor JSON.
- Return a direct remediation code and message for each failure.
- Reserve BACKEND_PROBE_UNAVAILABLE for failures that happen before trace readback can be classified.

## Why

Doctor previously collapsed most probe failures into BACKEND_PROBE_UNAVAILABLE. A user could not tell whether the project key failed, the trace was still processing, the pipeline reported a terminal failure, or the read API returned an invalid response.

The new result identifies the failed boundary without exposing response bodies, credentials, URLs, or server exception text.

## Compatibility

- Uses the existing POST /v1/traces ingestion route.
- Uses the existing GET /api/traces/v3/:traceId read route.
- Adds no endpoint, store, migration, or telemetry path.
- Keeps the Doctor v2 result contract and safe primitive details.

## Validation

- python -m neatlogs doctor --local --json: passed with four captured spans and successful flush.
- tests/unit/test_doctor_v2.py: 35 passed.
- Full Python suite: 454 passed.

### DIFF
--- a/neatlogs/doctor_v2.py
+++ b/neatlogs/doctor_v2.py
@@ -64,6 +64,11 @@ _REMEDIATION = {
     "CREDENTIAL_MISSING": "SET_CREDENTIAL",
     "AUTH_FAILED": "CHECK_INGEST_CREDENTIAL",
     "BACKEND_PROBE_UNAVAILABLE": "CHECK_TRACE_ENDPOINT",
+    "TRACE_READBACK_TIMEOUT": "WAIT_FOR_TRACE",
+    "INGESTION_PIPELINE_FAILED": "CONTACT_SUPPORT",
+    "BACKEND_HTTP_ERROR": "CHECK_TRACE_ENDPOINT",
+    "BACKEND_CONNECTION_FAILED": "CHECK_TRACE_ENDPOINT",
+    "TRACE_READBACK_INVALID": "CONTACT_SUPPORT",
     "ENDPOINT_INVALID": "SET_ENDPOINT",
     "PROVIDER_OWNERSHIP_AMBIGUOUS": "USE_PRIVATE_PROVIDER",
     "TRACE_ID_INVALID": "RECREATE_TRACE",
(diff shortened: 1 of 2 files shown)


neatlogs/doctor_v2.py
```diff
@@ -1204,62 +1221,133 @@ def doctor_probe_v2(
...
+                        raise _ProbeReadbackError(
+                            "TRACE_READBACK_INVALID",
+                            "Trace read-back returned an invalid terminal receipt",
+                        ) from exc
                     value = None
                 current_diagnostics = _ingestion_diagnostic_details(value)
                 if response.status_code == 409:
-                    last_diagnostics = current_diagnostics
-                    response.raise_for_status()
+                    if (
```
tanisha-neatlogs: [P1] Handle the backend’s valid DLQ response. GET /api/traces/v3/:traceId explicitly permits a 409 with finalizationStatus="dlq" and no ingestionDiagnostics. This branch classifies that valid terminal pipeline result as TRACE_READBACK_INVALID, so Doctor still misdiagnoses an actual ingestion failure whenever the optional receipt is unavailable—the exact failure mode these PRs are addressing. Please classify a schema-valid terminal 409 as INGESTION_PIPELINE_FAILED and attach the allowlisted receipt fields only when they are present. Add a regression using the backend DLQ body without ingestionDiagnostics.

tanisha-neatlogs (changes requested):
Requesting changes for the valid terminal-response classification below.

neatlogs/doctor_v2.py
shyam-neatlogs: Addressed in 844bd64. Doctor now accepts the backend's schema-valid `409` DLQ response without optional `ingestionDiagnostics` and reports `INGESTION_PIPELINE_FAILED`; it attaches details only when the allowlisted diagnostics parser accepts them. Added regressions for the valid no-diagnostics response and malformed DLQ bodies. Local checks and the full GitHub CI matrix, including packed-wheel E2E, are green.

tanisha-neatlogs (approved):

PR #97, opened 2026-09-07
Propagate managed prompt identity to provider LLM spans

niketan-neatlogs:
## Summary

`neatlogs.trace(name=..., system_prompt_template=...)` already carries prompt template content, variables, and version to auto-instrumented child LLM spans, but it dropped the stable prompt identity. Generic provider span names then caused the backend to group independent prompt families together.

This change propagates the trimmed prompt-bearing trace name through the existing cross-SDK `neatlogs.llm.prompt_key` contract. It introduces no new public SDK argument or Python-only wire attribute.

## What changed

- Carry the prompt key in the existing OTel prompt context only when a system template is supplied.
- Emit `neatlogs.llm.prompt_key` only on LLM descendants that receive that template.
- Preserve an identity already supplied by the instrumented LLM span through either `neatlogs.llm.prompt_key` or `traceloop.prompt.key`.
- Preserve nested scope restoration, async/thread isolation, exception cleanup, streaming span lifetime, deprecated `prompt_template=` support, and current auto-root behavior.
- Leave general trace scopes, user-template-only scopes, and non-LLM spans without an exported prompt identity.

## Impact

- **SDK users:** no API change; existing instrumentation continues to work.
- **Backend:** companion [neatlogs-app PR #1326](https://github.com/neatlogs/neatlogs-app/pull/1326) consumes the established canonical key and retains fallback behavior for older SDKs.
- **Data:** no schema migration or client-side database identifier.
- **Security/privacy:** only the existing trace scope name is used as prompt identity; prompt content handling is unchanged.
- **Cross-SDK contract:** uses the canonical attribute already shared by Python, TypeScript, and Go.

## Verification

- Full current SDK unit suite: 498 passed, 3 optional skips.
- Prompt identity suite: 14 cases, including existing-key precedence, normalization, nesting, empty names, exceptions, auto-root, async tasks, threads, and streaming.
- Prompt suite passed on Python 3.10, 3.11, 3.12, and 3.13.
- Black and isort checks passed for the changed Python files.
- Source distribution and wheel built successfully.
- Built-wheel prompt-contract smoke passed on Python 3.10 and 3.13.

## Rollout

Merge and deploy the backward-compatible backend consumer first. Release this SDK second, then run the dev-only Google GenAI multi-agent canary before customer adoption. An older backend ignores the additional canonical attribute, while an older SDK continues to use the backend fallback.

Tracks [NL-1776](https://linear.app/neatlogs/issue/NL-1776).

### DIFF
--- a/neatlogs/config/attribute-mapping.json
+++ b/neatlogs/config/attribute-mapping.json
@@ -313,13 +313,6 @@
             "target": "neatlogs.llm.response_id"
           },
           "prompt_template": {
-            "name": {
-              "sources": [
-                "neatlogs.llm.prompt_template.name",
-                "llm.prompt_template.name"
-              ],
-              "target": "neatlogs.llm.prompt_template.name"
-            },
             "template": {
               "sources": [
                 "neatlogs.prompt.template",
--- a/neatlogs/config/attribute-mapping.json
+++ b/neatlogs/config/attribute-mapping.json
@@ -313,6 +313,13 @@
             "target": "neatlogs.llm.response_id"
           },
           "prompt_template": {
+            "name": {
+              "sources": [
+                "neatlogs.llm.prompt_template.name",
+                "llm.prompt_template.name"
+              ],
+              "target": "neatlogs.llm.prompt_template.name"
+            },
             "template": {
               "sources": [
                 "neatlogs.prompt.template",
--- a/neatlogs/core/context.py
+++ b/neatlogs/core/context.py
@@ -220,6 +220,7 @@ def trace(
             c = set_value("neatlogs.system_prompt_variables", variables_json, context=c)
         if template_string:
             c = set_value("neatlogs.system_prompt_template", template_string, context=c)
+            c = set_value("neatlogs.system_prompt_name", name.strip(), context=c)
         if user_variables_json:
             c = set_value("neatlogs.user_prompt_variables", user_variables_json, context=c)
         if user_template_string:
--- a/neatlogs/core/context.py
+++ b/neatlogs/core/context.py
@@ -220,7 +220,7 @@ def trace(
             c = set_value("neatlogs.system_prompt_variables", variables_json, context=c)
         if template_string:
             c = set_value("neatlogs.system_prompt_template", template_string, context=c)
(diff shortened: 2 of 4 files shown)


neatlogs/config/attribute-mapping.json
```diff
@@ -313,6 +313,13 @@
             "target": "neatlogs.llm.response_id"
           },
           "prompt_template": {
+            "name": {
+              "sources": [
+                "neatlogs.llm.prompt_template.name",
+                "llm.prompt_template.name"
```
tanisha-neatlogs: **Blocker — use the existing canonical prompt identity.** `llm.prompt_template.name` is not an OpenInference attribute, and `neatlogs.llm.prompt_template.name` is absent from the TypeScript and Go canonical maps. All three SDKs already normalize the stable prompt identity as `neatlogs.llm.prompt_key` (including `traceloop.prompt.key`). Adding a Python-only wire key makes the same concept language-dependent and makes backend #1326 work only for this SDK. Please emit/use the existing canonical key, or first define a genuinely distinct field in the shared cross-SDK contract and update every SDK/docs consumer together.

tanisha-neatlogs (changes requested):
The propagation and concurrency tests are solid, and I found no regression in the effective merge with the recent HTTP, ADK, Pydantic AI, or Strands fixes. One wire-contract blocker remains inline: use the existing cross-SDK canonical prompt identity instead of adding a Python-only attribute.

neatlogs/config/attribute-mapping.json
niketan-neatlogs: Fixed in 0a9f6f3. The SDK no longer defines or emits llm.prompt_template.name / neatlogs.llm.prompt_template.name. Prompt-bearing trace context now emits the established neatlogs.llm.prompt_key, while an existing neatlogs.llm.prompt_key or traceloop.prompt.key on the LLM span takes precedence. The 14-case suite passes on Python 3.10-3.13; the full suite is 498 passed, and built-wheel smoke passed on 3.10/3.13.

tanisha-neatlogs (approved):

PR #98, opened 2026-09-08
Stop emitting HTTP transport spans

tanisha-neatlogs:
## Summary
- remove requests/httpx/urllib3/aiohttp span instrumentation from the public registry and dependencies
- reject HTTP instrumentation keys and manually authored HTTP trace kinds
- suppress HTTP transport spans before normalization, logs, root I/O backfill, masking, completion markers, and OTLP export
- preserve semantic AI spans that merely contain HTTP metadata
- update affected examples and regression coverage

The HTTP schema mapping remains for external-ingest compatibility; this PR only prevents the Python SDK from sending HTTP spans.

## Validation
- 481 unit tests passed; 3 skipped


niketan-neatlogs (commented):
Reviewed with audit-pr and thermo-nuclear-code-quality-review at `2afceb95dfc9f91733c2f2dfabdb7c3d4a4f279f` against `362989e84c561b6d6906b4914b633ada77770b58`.

3 actionable P2 findings posted inline. Python unit suite: 481 passed, 3 skipped. Three reviewer probes fail on this head and pass on the recorded base; they exercise the real init/processor/exporter pipeline.

The structural issues are described with the affected behavior and focused fixes in the inline comments. Review only; no fixes, merge or deployment performed.

neatlogs/core/span_processor.py
```diff
@@ -70,68 +65,56 @@ def _is_neatlogs_scope_span(span: Any) -> bool:
...
     try:
-        scope = getattr(span, "instrumentation_scope", None)
-        scope_name = getattr(scope, "name", "") or ""
-        if not scope_name.startswith(_HTTP_INSTRUMENTATION_SCOPES):
+        attrs = span.attributes or {}
+        neatlogs_kind = str(attrs.get("neatlogs.span.kind") or "").upper()
+        openinference_kind = str(attrs.get("openinference.span.kind") or "").upper()
+        if neatlogs_kind == "HTTP":
+            return True
+        if neatlogs_kind in _SEMANTIC_SPAN_KINDS:
```
niketan-neatlogs: [P2] Make normalization honor the same semantic-kind precedence

A span with `neatlogs.span.kind=RERANKER` and `openinference.span.kind=HTTP` is explicitly preserved by this helper, but the normalizer still gives the OpenInference kind precedence and rewrites the canonical kind to `http`. The new exporter filter then silently drops that semantic span. I reproduced this through `neatlogs.init()`, the real processor and exporter, with a nested RERANKER under a WORKFLOW: base exports the child (already mislabeled `http`), while this head exports only the root and completion marker. The new direct helper test does not exercise that normalization boundary. Use one canonical kind-resolution policy in both normalization and suppression, and add an end-to-end processor/exporter regression test that asserts the semantic kind and exported row.

neatlogs/init.py
```diff
@@ -638,10 +655,11 @@ def force_flush(self, timeout_millis: int = 30000):
             _media_store,
             diagnostics=_delivery_diagnostics,
         )
+        filtered_limited_span_exporter = _FilteredOTLPExporter(limited_span_exporter)
         batch_processor = ObservableBatchSpanProcessor(
             _FilteredOTLPExporter(
                 MaskingSpanExporter(
-                    limited_span_exporter,
+                    filtered_limited_span_exporter,
```
niketan-neatlogs, line 651: [P2] Gate completion on the post-mask root acceptance decision

When a supported mask callback changes a WORKFLOW root's kind to HTTP, the new inner filter suppresses that root after `_RootCompletionMarkerProcessor.on_end()` has already emitted `neatlogs.trace.complete`. A root with an ordinary TOOL child consequently exports `[child, completion marker]` on this head, versus `[child, root, completion marker]` on base. With the retained child, the backend accepts the completion marker and queues finalization, but cannot find the suppressed root and enters `retry-no-root`. Have the completion lifecycle consume the final post-mask acceptance result, rather than deciding before the asynchronous export filter; cover a masked root with a retained child.

neatlogs/init.py
```diff
@@ -616,7 +633,7 @@ def __init__(self, inner):
             def export(self, spans):
                 from opentelemetry.sdk.trace.export import SpanExportResult
 
-                kept = [s for s in spans if not is_rootless_infra_http(s)]
+                kept = [s for s in spans if not is_http_span(s)]
```
niketan-neatlogs: [P2] Release staged media when suppressing a span

This list comprehension discards spans without releasing their pending-media references. For an LLM span containing staged image bytes whose mask changes only its kind to HTTP, masking sees unchanged media tokens and releases nothing, while this inner filter prevents `TypedMediaSpanExporter` from consuming them. A local real-pipeline probe retains 1 item / 100,008 bytes after flush on this head; base releases both. Repeated suppressed spans fill the bounded pending-media store and prevent later valid spans from staging media until shutdown. Route both HTTP rejection boundaries through the existing reference-aware media cleanup using the rejected span snapshot (attributes and events), and test that suppression restores the media budget.

tanisha-neatlogs: Fixed in 29cd695. Both HTTP rejection boundaries release staged media from span attributes and events; the regression verifies the pending-media budget returns to zero after post-mask suppression.

neatlogs/init.py
tanisha-neatlogs: Fixed in 29cd695. Completion eligibility is now consumed only after the final post-mask filter accepts and successfully exports the root. A masked HTTP root with a retained TOOL child exports no completion marker.

neatlogs/core/span_processor.py
tanisha-neatlogs: Fixed in 29cd695. Normalization and HTTP filtering now share one NeatLogs-first canonical kind resolver, and the regression runs through the real processor/exporter boundary and asserts RERANKER is retained.

niketan-neatlogs (approved):

PR #99, opened 2026-09-08
fix: support current Pydantic AI releases

harshk-neatlogs:
## Summary

Pydantic AI 2.x changed framework-owned span names and removed the constructor-level instrumentation flag. The tests now assert stable semantic kinds and ancestry while keeping strict checks for duplicate spans, provider isolation, lifecycle behavior, errors, and cancellation.

The optional extra now installs `pydantic-ai-slim`, which avoids pulling provider packages that conflict with the SDK's existing Google GenAI dependency. CI covers Pydantic AI 1.44 and 2.41 with both the minimum and current OpenInference adapter lines from built wheels.

## Tests

- Full suite: 516 passed, 2 existing skips
- Four wheel compatibility lanes: 17 passed in each
- Wheel consumers on Python 3.10, 3.11, 3.12, and 3.13
- `pip check`, Black, isort, lock check, wheel build, and source build

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,8 +152,8 @@ portkey = [
     "portkey-ai>=1.0.0",
 ]
 pydantic-ai = [
-    "pydantic-ai>=0.7.5",
-    "openinference-instrumentation-pydantic-ai[instruments]>=0.1.20",
+    "pydantic-ai-slim>=1.44.0",
+    "openinference-instrumentation-pydantic-ai>=0.1.20",
 ]
 smolagents = [
     "smolagents>=1.0.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,7 +152,7 @@ portkey = [
     "portkey-ai>=1.0.0",
 ]
 pydantic-ai = [
-    "pydantic-ai-slim>=1.44.0",
+    "pydantic-ai>=0.7.5",
     "openinference-instrumentation-pydantic-ai>=0.1.20",
 ]
 smolagents = [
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -152,8 +152,8 @@ portkey = [
     "portkey-ai>=1.0.0",
 ]
 pydantic-ai = [
-    "pydantic-ai>=0.7.5",
-    "openinference-instrumentation-pydantic-ai[instruments]>=0.1.20",
+    "pydantic-ai-slim>=1.44.0",
```
tanisha-neatlogs: **Blocker — this narrows the public extra instead of only adding 2.x compatibility.** `pip install "neatlogs[pydantic-ai]"` currently installs `pydantic-ai>=0.7.5`, whose normal distribution includes its provider integrations. Bare `pydantic-ai-slim` does not; for example, the shipped `examples/sdk_examples/pydantic_ai_basic.py` uses `OpenAIProvider`, but this revised extra no longer installs `openai`. The compatibility job masks that packaging regression because `.github/requirements-test.txt` pulls in `openai` through AutoGen. The new `>=1.44.0` floor also silently drops the advertised 0.7.5–1.43 range even though adapter 0.1.20 supports `pydantic-ai>=0.7.5`. Please preserve the existing install/minimum-version contract (the adapter can still lose its `[instruments]` extra) and add the old minimum plus current 2.x to the wheel matrix.

tanisha-neatlogs (changes requested):
The compatibility tests are directionally good, but the public extra currently regresses both its provider-install contract and its supported version range. The blocking detail is inline.

pyproject.toml
harshk-neatlogs: fixed - extra is back to full pydantic-ai>=0.7.5 so the providers install again, adapter stays without [instruments]. wheel matrix now covers 0.7.5, 1.44.0 and 2.41.0 against adapters 0.1.20/0.1.25. two wrinkles: 0.7.x needs otel <1.44 (opentelemetry._events got removed in 1.44) so those lanes pin 1.43.0, and the 2.x provider ranges clash with the shared openai/google-genai test pins so the job uses a trimmed requirements file (the test doesn't import either). verified from the built wheel on all three versions - pip check clean, 17/17 semantic tests pass on each

tanisha-neatlogs (approved):

PR #117, opened 2026-09-15
fix: end Pydantic AI spans on cancellation

### LINKED ISSUE #116: Pydantic AI cancellation leaves AGENT and model spans open until shutdown
## Summary

Cancelling a manually wrapped Pydantic AI `Agent.run()` leaves its AGENT span unfinished until provider shutdown. The cancellation also reaches the underlying model span, so both stay open and are later force-ended with `UNSET` status.

## Reproduction

```python
task = asyncio.create_task(neatlogs.wrap(agent).run("wait forever"))
await started.wait()
task.cancel()
with pytest.raises(asyncio.CancelledError):
    await task
```

Run twice on `main` at `1a8292f45c738f7a78d870b30f4c7829a2f743f2`. Both normalized outputs were identical:

```text
finished_before_shutdown []
Ended 2 active Neatlogs span(s) during cancelled
finished_after_shutdown [('pydantic_ai.model.request', 'UNSET', True), ('pydantic_ai.agent.run', 'UNSET', True)]
```

## Cause

The async wrapper catches `Exception`, but `asyncio.CancelledError` inherits from `BaseException`. Cancellation bypasses `_err(span, e)`, then the function exits before `_finalize_run_span()`.

## Expected

Cancellation should still propagate, while the AGENT span ends immediately with `ERROR` status instead of remaining open until shutdown.

harshk-neatlogs:
Fixes #116

Fixes the Pydantic AI cancellation span leak.

The async manual wrapper now handles `BaseException`, which includes `asyncio.CancelledError`, records the span error, ends the span, and re-raises cancellation unchanged.

Tests:
- cancellation repro: 20/20 normalized runs identical
- Pydantic AI unit suite: 18 passed


neatlogs/pydantic_ai.py
```diff
@@ -190,7 +190,10 @@ async def patched_run(*args, **kwargs):
         start = time.perf_counter()
         try:
             result = await orig_run(*args, **kwargs)
-        except Exception as e:
+        except BaseException as e:
```
tanisha-neatlogs: [P2] Please handle asyncio.CancelledError separately before the ordinary error path. Passing it to _err marks normal task cancellation ERROR and records it as an application exception, while the existing stream and decorator paths mark cancellation and use StatusCode.UNSET. Catch CancelledError, end the span with cancellation semantics, and re-raise; keep except Exception for actual failures. This also avoids routing KeyboardInterrupt, SystemExit, and GeneratorExit through _err.

tanisha-neatlogs (changes requested):
Cancellation should close the span without being reported as an application failure.

tanisha-neatlogs (approved):
Re-reviewed the current head. The async manual wrapper now handles asyncio.CancelledError separately, marks cancellation, ends the span with UNSET status, and re-raises. The focused cancellation regression passes. The unrelated native tool input.value test failure also reproduces on main with my local dependency set.

PR #121, opened 2026-09-16
fix: preserve evaluator kind in canonical telemetry

### LINKED ISSUE #120: Canonical telemetry v2 exports EVALUATOR spans as UNKNOWN
## Summary

Python's public tracing and decorator APIs accept `EVALUATOR` spans, but canonical telemetry v2 silently exports them as `UNKNOWN`.

`neatlogs/core/telemetry_v2.py` omits `EVALUATOR` from `_KINDS`, and the packaged Python and Go canonical schemas omit it too. The TypeScript canonical schema already includes it.

## Reproduction

Reproduced twice on published Python SDK 1.4.23 and twice on current `main`:

```json
{"emitted_kind":"UNKNOWN","schema_validation":"accepted","sdk":"1.4.23"}
```

Expected: `emitted_kind=EVALUATOR`

Actual: `emitted_kind=UNKNOWN`

This loses evaluator identity in doctor and canonical exports even though the span itself is created correctly.

## Fix

- include `EVALUATOR` in the Python canonical normalizer
- align the Python and Go canonical schemas with TypeScript
- include the evaluator semantic namespace
- update the embedded schema digests
- add regression coverage in both SDKs

Validation:

- Python focused schema/code/pipeline suite: 37 tests passed, run twice
- Go: `go test ./contract/v2` passes

harshk-neatlogs:
## Summary

Preserves `EVALUATOR` in canonical telemetry v2 instead of normalizing it to `UNKNOWN`.

- adds `EVALUATOR` to the normalizer and canonical schema
- adds the evaluator semantic namespace
- updates the schema digest
- adds `test_canonical_telemetry_preserves_evaluator_kind`

## Validation

Focused schema/code/pipeline suite: 37 tests passed, run twice.

Fixes #120

The matching Go contract alignment is filed separately and linked from this issue.


neatlogs/contracts/v2/neatlogs-telemetry.schema.json
```diff
@@ -322,6 +322,7 @@
         "VECTOR_STORE",
         "MEMORY",
         "GUARDRAIL",
+        "EVALUATOR",
```
tanisha-neatlogs, line 325: [P2] Please publish this contract change from the declared authority first. As of current neatlogs/skills main, contracts/v2/manifest.json publishes SHA 9aec0e1b4e2fba718a1bad060798a881543c56ec8b887c6b0fb8ab147bbaee75 and its schema does not include EVALUATOR or neatlogs.evaluator.*; this PR ships ae79717d127e2faa761ad83c8a51a08cae6fce00a04446d46dcc6d1a371818ce while still calling Skills the authority and keeping contract version 2.0.0. That gives the same canonical version different bytes across repos. Land the Skills contract change first, then regenerate and sync the SDK copies.

tanisha-neatlogs (changes requested):
The evaluator change needs to come from the declared canonical contract before the SDK copies are updated.

PR #131, opened 2026-09-20
fix(crewai): end async batch span on cancellation

### LINKED ISSUE #130: CrewAI kickoff_for_each_async leaves span open on cancellation
## What happened

Cancelling an in-flight `Crew.kickoff_for_each_async()` call propagates `asyncio.CancelledError`, but the Neatlogs workflow span stays open and keeps `UNSET` status.

This leaves an unfinished span whenever application shutdown, a timeout, or caller cancellation interrupts a CrewAI batch.

## Reproduction

Tested with:

- `neatlogs==1.4.23`
- Python 3.10.12
- `opentelemetry-api==1.44.0`
- `opentelemetry-sdk==1.44.0`

```python
import asyncio

from neatlogs import crewai as instrumentation


class Crew:
    async def kickoff_for_each_async(self, inputs):
        await asyncio.Event().wait()


async def main():
    instrumentation._patch_crew_class(Crew)
    task = asyncio.create_task(Crew().kickoff_for_each_async([{"x": 1}]))
    await asyncio.sleep(0)
    task.cancel()

    try:
        await task
    except asyncio.CancelledError:
        pass


asyncio.run(main())
```

I ran the isolated repro twice with identical normalized output:

```text
{"package":"neatlogs==1.4.23","spans":[{"ended":false,"name":"crewai.crew.kickoff_for_each_async","status":"UNSET"}]}
```

Proof hash for both runs:

```text
417eaa2ed07b714147b53b26e8aff3d177a9fef1ad97b8fb2003d2b9274c40f6
```

## Cause

The async batch wrapper catches `Exception`, but `asyncio.CancelledError` inherits directly from `BaseException` on current Python versions. Cancellation skips `_err(...)`, so the span is never marked as an error or ended.

## Expected behavior

Cancellation should still propagate to the caller, while the instrumentation marks the span as `ERROR` and ends it.

harshk-neatlogs:
## Summary

- handle `CancelledError` in CrewAI's `kickoff_for_each_async` instrumentation
- mark the workflow span as `ERROR` and end it before cancellation propagates
- add a focused async cancellation regression

Fixes #130

## Verification

- published `neatlogs==1.4.23` repro: 2 identical runs
- before: span stays open with `UNSET` status
- fixed repro: 2 identical runs, span ends with `ERROR` status
- focused regression: `1 passed`


harshk-neatlogs:
kickoff_for_each_async now uses the same CancelledError branch: cancelled attr + UNSET, and the exception isn't recorded. Error path still ERROR, with a test. 2/2 pass. Heads up, #134 adds the same _cancelled helper so whichever merges second will need a small rebase.

PR #134, opened 2026-09-20
fix(crewai): end agent and flow spans on cancellation

### LINKED ISSUE #133: CrewAI agent and flow async cancellation leave root spans open
## What happened

Cancelling standalone CrewAI `Agent.kickoff_async()` or `Flow.kickoff_async()` calls propagates `asyncio.CancelledError`, but Neatlogs leaves the root span open with `UNSET` status.

This is separate from #130. That one covers `Crew.kickoff_for_each_async`; these two wrappers have their own cancellation paths.

## Reproduction

Tested with:

- `neatlogs==1.4.23`
- `crewai==1.15.22`
- Python 3.10.12
- `opentelemetry-sdk==1.44.0`

Both isolated repros were run twice with identical normalized output.

Agent:

```text
{"package":"neatlogs==1.4.23","spans":[{"ended":false,"name":"crewai.agent.demo","status":"UNSET"}]}
b564e3d431c10b40d02e0b1d6755173bfb112d471e40c8758e9550c4d89adfff
```

Flow:

```text
{"package":"neatlogs==1.4.23","spans":[{"ended":false,"name":"crewai.flow.kickoff_async","status":"UNSET"}]}
485be48054a034f15965d5750f0ad9b9c59758e6352029e0c76424ff7a987c53
```

## Cause

The native async Agent and Flow wrappers catch `Exception`, but `asyncio.CancelledError` inherits directly from `BaseException`. Cancellation skips `_err(...)`, so both spans stay open.

## Expected behavior

Cancellation should still propagate, while both root spans are ended and marked `ERROR`, matching the cancellation handling in #131.

harshk-neatlogs:
## Summary

- catch `BaseException` in the native async Agent and Flow wrappers so `asyncio.CancelledError` reaches cleanup
- mark the root span `ERROR`, record the cancellation, end the span, then re-raise unchanged
- add focused regressions for both wrappers

Fixes #133.

## Test plan

```text
pytest -q tests/unit/test_crewai_agent_flow_async_cancellation.py
2 passed
```

Agent repro SHA-256: `b564e3d431c10b40d02e0b1d6755173bfb112d471e40c8758e9550c4d89adfff`

Flow repro SHA-256: `485be48054a034f15965d5750f0ad9b9c59758e6352029e0c76424ff7a987c53`


harshk-neatlogs:
Agent and flow kickoff_async now handle CancelledError separately (cancelled attr + UNSET, nothing recorded as an exception) and _err is back to Exception. Added an error-path test for flow. 3/3 pass, full suite same as main.

PR #136, opened 2026-09-20
fix(openai): end async spans on cancellation

### LINKED ISSUE #135: OpenAI async cancellation leaves chat and Responses spans open
## What happened

Cancelling either native OpenAI async entry point propagates `asyncio.CancelledError`, but Neatlogs leaves the root span open with `UNSET` status:

- `AsyncOpenAI().chat.completions.create(...)`
- `AsyncOpenAI().responses.create(...)`

## Reproduction

Tested with `neatlogs==1.4.23` on Python 3.10.12. Both isolated repros were run twice with identical normalized output.

Async chat completions:

```text
{"package":"neatlogs==1.4.23","spans":[{"ended":false,"events":[],"name":"openai.chat.completions.create","status":"UNSET"}]}
3206cf56073e30a5da88bbe9bf4354c1744808a9f956bd0a5ef6915a15eef0b7
```

Async Responses API:

```text
{"package":"neatlogs==1.4.23","spans":[{"ended":false,"events":[],"name":"openai.responses.create","status":"UNSET"}]}
ee80eac53691b55bbb5adc9cc3e7604586a91a3636be9dacc627b476d4d32a71
```

Each hash above was identical across both isolated runs.

## Cause

Both native async wrappers catch `Exception`, but `asyncio.CancelledError` inherits directly from `BaseException`. Cancellation skips span cleanup.

## Expected behavior

Cancellation should still propagate unchanged, while the root span records the cancellation, ends, and is marked `ERROR`.

harshk-neatlogs:
## Summary

- catch `BaseException` in native OpenAI async chat completions and Responses wrappers so `asyncio.CancelledError` reaches cleanup
- record the cancellation, mark the root span `ERROR`, end it, and re-raise unchanged
- add focused regressions for both async entry points

Fixes #135.

## Reproduction

Published `neatlogs==1.4.23`, two identical isolated runs per API:

- chat completions: `3206cf56073e30a5da88bbe9bf4354c1744808a9f956bd0a5ef6915a15eef0b7`
- Responses API: `ee80eac53691b55bbb5adc9cc3e7604586a91a3636be9dacc627b476d4d32a71`

Fixed branch, two identical isolated runs per API:

- chat completions: `55d8abeb3fb169d76a1f076c607f14a2d57c1cc0eb7e51a34a316f9c972ae8db`
- Responses API: `03c7e6932da6372711295d6b79d89f2d4406b4687c5858cc3afc285d47a11183`

## Test plan

```text
pytest -q tests/unit/test_openai_async_cancellation.py
2 passed
```


harshk-neatlogs:
Moved cancellation to its own CancelledError branch, same as #117. It sets neatlogs.stream.cancelled, leaves status UNSET and re-raises, so CancelledError is no longer recorded as an exception. Provider errors still go through the Exception path as ERROR. Added a test for that. Unit tests 4/4, full suite same as main.

Issue #138, opened 2026-09-21
Canonical telemetry v2 drops captured LLM tool definitions


harshk-neatlogs:
## Problem

The provider wrappers capture LLM tool definitions under `neatlogs.llm.tools.{i}.*`, but canonical telemetry v2 hardcodes `semantic.request.tools` to an empty list.

This drops the tool name, description and input schema from canonical output even though the span contains them.

The affected path is `neatlogs/core/telemetry_v2.py` in `_llm_semantic`.

## Reproduction

Tested on Python SDK `1.4.23` / repository commit `e50f8a5873f77a38380d35ce5d08ea8710e542bb` with an OpenAI-style `get_weather` tool definition.

Run 1:

```text
[]
4f53cda18c2baa0c0354bb5f9a3ecbe5ed12ab4d8e11ba873c2f11161202b945
```

Run 2:

```text
[]
4f53cda18c2baa0c0354bb5f9a3ecbe5ed12ab4d8e11ba873c2f11161202b945
```

The complete raw output files were identical:

```text
0b9952ffb509569fb9f529702ccbf7ef28cf0b075fb2cc8853b3612fd6c2eafc  canonical-tools-run1.txt
0b9952ffb509569fb9f529702ccbf7ef28cf0b075fb2cc8853b3612fd6c2eafc  canonical-tools-run2.txt
```

The verified hosted screenshot shows the live source at `e50f8a5` hardcoding `"tools": []`: https://files.instinct.com/file-01M310PJ3D0ZGYPCBF53A3HFH5

## Expected

Canonical telemetry should preserve each valid wrapper-captured tool definition, including:

- type
- name
- description
- input schema
- provider configuration when present

Malformed records without a tool name should be skipped.

## Impact

Any canonical telemetry consumer sees no available tools even when the wrapper captured them correctly. This removes useful context for tool-call debugging and makes the canonical record disagree with the source span.

## Validation plan

Add a focused regression that emits an OpenAI-style tool definition and checks the canonical request. The fix should preserve the parsed schema and keep invalid nameless records out.

PR #139, opened 2026-09-21
fix(telemetry): preserve LLM tool definitions

harshk-neatlogs:
## Summary

- build canonical `semantic.request.tools` from `neatlogs.llm.tools.{i}.*` attributes
- preserve type, name, description, parsed input schema and provider configuration
- skip malformed tool records without a name
- add focused canonical telemetry coverage

## Problem

All major provider wrappers capture LLM tool definitions, but telemetry v2 hardcodes the canonical request `tools` field to an empty list. The tool name, description and schema disappear from canonical output.

Before the fix, two isolated runs returned:

```text
[]
4f53cda18c2baa0c0354bb5f9a3ecbe5ed12ab4d8e11ba873c2f11161202b945
```

The complete raw run-file hash also matched twice:

```text
0b9952ffb509569fb9f529702ccbf7ef28cf0b075fb2cc8853b3612fd6c2eafc
```

After the fix, both runs returned the preserved `get_weather` definition with the same normalized hash:

```text
579f37795dd572dcddbeeeebf10c587997c5ec43384d35d454cda9a713a36e65
```

## Validation

```text
6 passed in 0.26s
```

```text
black: 2 files would be left unchanged
```

Related: #138


harshk-neatlogs:
Type-only tools like web_search now use the type as the name instead of being dropped. Records with no name and no type are skipped. The new test also validates the envelope against the schema. 12/12 pass.

